PR #24, opened 2021-05-12
Update sample data naming conventions

### LINKED ISSUE #3: Update sample detectors names & indices
Currently, sample detectors & their corresponding indices used to detect over are prefixed with `opendistro-`. This prefix will need to be removed.

Note this will involve updating logic used to determine the existence of created sample detectors as well.

Will need to consider backwards compatibility to accommodate users who have already created detectors/indices with the `opendistro-` prefix.

ohltyler:
### Description
Current sample detectors and indices include the legacy `opendistro-` prefix. This PR removes that prefix.

Example: 
eCommerce detector change: `opendistro-sample-ecommerce-detector` => `sample-ecommerce-detector`
eCommerce index change: `opendistro-sample-ecommerce` => `sample-ecommerce`

In addition:
- Handled backwards compatibility by ensuring legacy detectors created are still recognized as created on the sample detectors page
- Changed the sample details flyout to pull the detector & index names from an existing detector if possible. That way, new and legacy detectors will both be properly reflected
- Added legacy checks to helper fns and query parameters to search for both new and legacy created detectors and indices
- Confirmed newly named detectors/indices function the same as before

Screenshots:

Legacy + new sample detectors coexisting:
Screen Shot 2021-05-12 at 1 24 19 PM
Screen Shot 2021-05-12 at 1 24 35 PM

Details flyout (legacy):
Screen Shot 2021-05-12 at 1 25 02 PM

Details flyout (new):
Screen Shot 2021-05-12 at 1 25 16 PM

### Issues Resolved
Resolves #3

### Check List

### DIFF
--- a/server/routes/sampleData.ts
+++ b/server/routes/sampleData.ts
@@ -73,7 +73,7 @@ export default class SampleDataService {
             __dirname,
             '../sampleData/rawData/httpResponses.json.gz'
           );
-          indexName = 'opendistro-sample-http-responses';
+          indexName = 'sample-http-responses';
           break;
         }
         case SAMPLE_TYPE.ECOMMERCE: {
@@ -81,7 +81,7 @@ export default class SampleDataService {
             __dirname,
             '../sampleData/rawData/ecommerce.json.gz'
           );
-          indexName = 'opendistro-sample-ecommerce';
+          indexName = 'sample-ecommerce';
           break;
         }
         case SAMPLE_TYPE.HOST_HEALTH: {
@@ -89,7 +89,7 @@ export default class SampleDataService {
             __dirname,
             '../sampleData/rawData/hostHealth.json.gz'
           );
-          indexName = 'opendistro-sample-host-health';
+          indexName = 'sample-host-health';
           break;
         }
       }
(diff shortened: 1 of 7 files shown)


kaituo (approved):

server/routes/sampleData.ts
```diff
@@ -73,23 +73,23 @@ export default class SampleDataService {
             __dirname,
             '../sampleData/rawData/httpResponses.json.gz'
           );
-          indexName = 'opendistro-sample-http-responses';
+          indexName = 'sample-http-responses';
```
ylwu-amzn, line 76: If user already has index with name `sample-http-responses`, will we create sample detector on user's existing index or will throw exception?

ohltyler: We will try to create on the existing index, but if it's some user's created index it will most likely fail since it won't have the same fields & mappings, etc.

Basically we just check for the existence of the sample index name [here](https://github.com/ohltyler/anomaly-detection-dashboards-plugin-1/blob/sample-rename/public/pages/SampleData/containers/SampleData/SampleData.tsx#L134) and create if it doesn't already exist. 

This check is added to handle the case of user creating the sample data (both detector and index), and maybe deleting or changing the detector name. In that case, they can recreate the sample detector, and will use the existing sample index if it is still there.

This is admittedly simple; one of the tracked improvements for sample detectors is adding delete buttons on the UI and better handling of edge cases

ylwu-amzn: Sure. Should pay attention when delete sample detectors to avoid delete some user's data index with the same name of `sample-http-responses`

ylwu-amzn (approved):
LGTM. Thanks for the change!

server/routes/sampleData.ts
ohltyler: Agreed. Will have to be careful when introducing deletions here. Same goes for detectors.

PR #36, opened 2021-06-01
Run CI against 1.0.0-rc1 docker images

ohltyler:

### Description

Changes the CI to run against the `1.0.0-rc1` staging docker images.

### Check List

### DIFF
--- a/.github/workflows/e2e-tests-workflow.yml
+++ b/.github/workflows/e2e-tests-workflow.yml
@@ -11,8 +11,7 @@ env:
   AD_OPENSEARCH_DASHBOARDS_PLUGIN_NAME: anomalyDetectionDashboards
   OPENSEARCH_DOCKER_IMAGE: opensearchstaging/opensearch
   DASHBOARDS_DOCKER_IMAGE: opensearchstaging/opensearch-dashboards
-  OPENSEARCH_DOCKER_TAG: 1.0.0-rc1
-  DASHBOARDS_DOCKER_TAG: 1.0.0-rc1
+  DOCKER_TAG: 1.0.0-rc1
 jobs:
   test-with-security:
     name: Run e2e tests with security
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-tests-workflow.yml
```diff
@@ -66,14 +68,14 @@ jobs:
...
           then
             ## Populate the Dockerfiles
-            echo "FROM $OPENSEARCH_DOCKER_IMAGE:latest" >> Dockerfile-AD-OpenSearch
-            echo "FROM $DASHBOARDS_DOCKER_IMAGE:latest" >> Dockerfile-AD-OpenSearch-Dashboards
+            echo "FROM $OPENSEARCH_DOCKER_IMAGE:$OPENSEARCH_DOCKER_TAG" >> Dockerfile-AD-OpenSearch
+            echo "FROM $DASHBOARDS_DOCKER_IMAGE:$DASHBOARDS_DOCKER_TAG" >> Dockerfile-AD-OpenSearch-Dashboards
             echo "COPY build/$AD_OPENSEARCH_DASHBOARDS_PLUGIN_NAME-$plugin_version.zip ." >> Dockerfile-AD-OpenSearch-Dashboards
             ## Uninstall existing AD artifact and install new one
-            echo "RUN if [ -d /usr/share/opensearch-dashboards/plugins/$AD_OPENSEARCH_DASHBOARDS_PLUGIN_NAME ]; then /usr/share/opensearch-dashboards/bin/opensearch-dashboards-plugin remove $AD_OPENSEARCH_DASHBOARDS_PLUGIN_NAME; fi" >> Dockerfile-AD-OpenSearch-Dashboards
+            echo "RUN if [ -d /usr/share/opensearch-dashboards/plugins/anomaly-detection-dashboards ]; then /usr/share/opensearch-dashboards/bin/opensearch-dashboards-plugin remove anomaly-detection-dashboards; fi" >> Dockerfile-AD-OpenSearch-Dashboards
```
ohltyler, line 78: This is due to the default plugin names being installed as kebab-case instead of camelCase in Dashboards ([PR](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/357)). Eventually, we will change the `AD_OPENSEARCH_DASHBOARDS_PLUGIN_NAME`, which is the zip name, to be consistent and kebab-case as well.

.github/workflows/e2e-tests-workflow.yml
```diff
@@ -66,14 +68,14 @@ jobs:
           plugin_version=$(node -pe "require('./package.json').version")
           echo plugin version: $plugin_version
           echo plugin name: $AD_OPENSEARCH_DASHBOARDS_PLUGIN_NAME
-          if docker pull $OPENSEARCH_DOCKER_IMAGE:latest
+          if docker pull $OPENSEARCH_DOCKER_IMAGE:$OPENSEARCH_DOCKER_TAG
```
ylwu-amzn, line 71: The `latest` will not point to `1.0.0-rc1`?

ohltyler: It's a good point, the digests do match on docker hub for `latest` and `1.0.0-rc1`. The reason I was thinking of explicitly setting it to `1.0.0-rc1`, was to treat it as pre-release testing, and just be more explicit/verbose. Whatever is tagged as `1.0.0-rc1` will be the one eventually promoted to the release docker image, and is the one we want to be sure our plugin is compatible with.

Let me know what you think.

ylwu-amzn: Got it, that's ok. We can move to latest later to make sure all test can pass on latest version.

ylwu-amzn: Not sure if backward compatibility is required for future OpenSearch versions. We may need to add multiple integration test workflows to run on old versions to test if latest code can work on all OpenSearch versions.

ohltyler: Since plugins are still required to have the same version as core, I think this is not a problem for now, since new plugin versions currently can't be installed on core versions that are mismatched.

ylwu-amzn (approved):

.github/workflows/e2e-tests-workflow.yml
```diff
@@ -11,6 +11,8 @@ env:
   AD_OPENSEARCH_DASHBOARDS_PLUGIN_NAME: anomalyDetectionDashboards
   OPENSEARCH_DOCKER_IMAGE: opensearchstaging/opensearch
   DASHBOARDS_DOCKER_IMAGE: opensearchstaging/opensearch-dashboards
+  OPENSEARCH_DOCKER_TAG: 1.0.0-rc1
+  DASHBOARDS_DOCKER_TAG: 1.0.0-rc1
```
kaituo, line 15: Can we assume those two are always the same?  If so, we can have just one tag.

ohltyler: sure - I'll combine into a single field.

kaituo (approved):

PR #66, opened 2021-07-29
Support multiple category fields

ohltyler:
### Description

Adds support for detectors to have multiple category fields. This is already supported in the backend plugin - see [here](https://github.com/opensearch-project/anomaly-detection/pull/121). Supports both real-time & historical detectors.

Main changes include:
- updates `CategoryField` component to support multiple selections in the combo box (up to 2)
- updates various chart components to show a list of selected category fields:
     1 - In the 'View by' title at the top of the anomalies chart
     2 - On y-axis (each field delimited by a newline)
     3 - When hovering over a heatmap cell. This involved adding a `customdata` field in the plot data in order to show another 
           line of text (see [here](https://plotly.com/python/hover-text-and-formatting/#adding-other-data-to-the-hover-with-customdata-and-a-hovertemplate) for more details)
     4- In chart titles
     5 - In anomaly table
- updates `AdditionalSettings` in detector config page to support showing multiple category fields
- corresponding unit tests

Note that the bulk of these changes are related to changing the key of a single category field value that was used before, to making it more generalizable and supporting a list of category fields as a key. This way, we can easily support 3,4,5, etc. category fields without making code changes.

Some changes had to be made to the queries to fetch anomaly results for high-cardinality detectors. Similar to what's mentioned in the paragraph above, the queries were usually contingent on using a single category field as a key for fetching data, which will fetch incorrect results in the multi-category-field scenario. To handle this, several queries had to change:
1. `getTopAnomalousEntitiesQuery`([link](https://github.com/ohltyler/anomaly-detection-dashboards-plugin-1/blob/0771e7a5f02a0ee40d7d9f9470776540ff54947a/public/pages/utils/anomalyResultUtils.ts#L945)): to support BWC, we still will fetch single-category-field results the same way as before, and perform a terms aggregation on a single category field. To support multi-category-field results, we use a new query that performs a terms aggregation on a new `model_id` field, such that each unique set of category field values will be mapped to a single model id. This was introduced in the backend plugin as part of support for multi-category detectors.
2. `getAnomalySummaryQuery` ([link](https://github.com/ohltyler/anomaly-detection-dashboards-plugin-1/blob/0771e7a5f02a0ee40d7d9f9470776540ff54947a/public/pages/utils/anomalyResultUtils.ts#L311)), `getBucketizedAnomalyResultsQuery` ([link](https://github.com/ohltyler/anomaly-detection-dashboards-plugin-1/blob/0771e7a5f02a0ee40d7d9f9470776540ff54947a/public/pages/utils/anomalyResultUtils.ts#L408)), `getEntityAnomalySummariesQuery` ([link](https://github.com/ohltyler/anomaly-detection-dashboards-plugin-1/blob/0771e7a5f02a0ee40d7d9f9470776540ff54947a/public/pages/utils/anomalyResultUtils.ts#L1168)): similar to `getTopAnomalousEntitiesQuery`, for these functions, the filtering has been factored out into a helper function `getEntityFilters` which is used to determine if the results should be filtered out by a category field (if only one category field), or model id (if multiple category fields)
3. `getAnomalyResults` ([link](https://github.com/ohltyler/anomaly-detection-dashboards-plugin-1/blob/0771e7a5f02a0ee40d7d9f9470776540ff54947a/server/routes/ad.ts#L677)), which is used for fetching the raw results when a heatmap cell is selected, now uses an optionally-passed-in entity list to dynamically generate filters for the category fields before making the call to OpenSearch. Note that `model_id` is not used here, since we don't store/know the individual model ID when making the call. Also, with future plans to support results using a subset of category fields, this implementation leaves it more flexible for future changes. 

Screenshots:

Category field component, with wording changes indicating up to 2 can now be selected:
Screen Shot 2021-08-11 at 3 47 08 PM

Entity list when hovering over heatmap cell: 
Screen Shot 2021-07-29 at 10 08 23 AM

Additional settings component:
Screen Shot 2021-07-29 at 10 09 17 AM

Full results page (changes highlighted in red):
screencapture-localhost-5601-app-anomaly-detection-dashboards-2021-07-29-10_08_38

### Check List


public/pages/utils/anomalyResultUtils.ts
```diff
@@ -992,64 +948,67 @@ export const getTopAnomalousEntitiesQuery = (
...
-    aggs: {
-      [TOP_ENTITIES_FIELD]: {
-        nested: {
-          path: ENTITY_FIELD,
         },
         aggs: {
           [TOP_ENTITY_AGGS]: {
             terms: {
-              field: ENTITY_VALUE_PATH_FIELD,
+              field: MODEL_ID_FIELD,
```
kaituo, line 991: Would this work for historical detector?  I don't recall @ylwu-amzn  has added model id field.

public/pages/ConfigureModel/components/CategoryField/CategoryField.tsx
```diff
@@ -144,24 +144,26 @@ export function CategoryField(props: CategoryFieldProps) {
...
-                        );
+                      const selection = options.map((option) => option.label);
+                      if (!isEmpty(selection)) {
+                        if (selection.length <= 2) {
+                          form.setFieldValue('categoryField', selection);
+                          form.setFieldValue(
+                            'shingleSize',
+                            MULTI_ENTITY_SHINGLE_SIZE
+                          );
+                        }
```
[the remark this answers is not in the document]
ohltyler: Current code will leave the original selected 2 fields, but will ignore any subsequent selected fields. For example, user selects [`a`, `b`, `c`, `d`], selection will be [`a`, `b`].

I agree some added wording around this behavior may be helpful, like some toast saying only 2 are supported currently. Will sync with UX

public/pages/AnomalyCharts/utils/anomalyChartUtils.tsx
```diff
@@ -413,11 +435,14 @@ export const getEntitytAnomaliesHeatmapData = (
   const timeStamps = plotTimes.map((timestamp) =>
     moment(timestamp).format(HEATMAP_X_AXIS_DATE_FORMAT)
   );
+  const entityListsTransformed = transformEntityListsForHeatmap(entityLists);
+
   const plotData = buildHeatmapPlotData(
     timeStamps,
-    entityValues.reverse(),
+    entityStrings.reverse(),
```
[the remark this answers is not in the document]
ohltyler: After doing some research, I believe it's related to the default orientation of the y-axis for Plotly heatmaps, so Yizhe had flipped the data as a workaround. ([reference](https://community.plotly.com/t/heatmap-y-axis-is-reversed-by-default-going-against-standard-convention-for-matrices/32180/2))

public/pages/utils/anomalyResultUtils.ts
ohltyler: I believe it does. I've tested on the test domain and historical multi-category results are returning model IDs. @ylwu-amzn can you confirm if this is part of what's included in [this PR](https://github.com/opensearch-project/anomaly-detection/pull/159)?

public/pages/ConfigureModel/components/CategoryField/CategoryField.tsx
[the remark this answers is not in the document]
kaituo: Got it.  Let me know what UX thinks.

public/pages/utils/anomalyResultUtils.ts
kaituo: just checked.  Seems Yaliang did add model id: https://github.com/opensearch-project/anomaly-detection/blob/f0e265e5f4ed45f5af491a7a43a17b735580289b/src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java#L1082

public/pages/ConfigureModel/components/CategoryField/CategoryField.tsx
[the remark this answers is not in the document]
ohltyler: Confirmed with UX & added some wording to specify that up to 2 category fields can be selected.

Commit: https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/66/commits/de4d6a9f3b827f9badd5cbfc8ce31b2be463ab94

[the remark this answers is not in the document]
kaituo: nice.

public/pages/AnomalyCharts/containers/AnomalyHeatmapChart.tsx
```diff
@@ -228,6 +229,23 @@ export const AnomalyHeatmapChart = React.memo(
       return false;
     };
 
+    // Custom hook to refresh all of the heatmap data when running a historical task
+    useEffect(() => {
+      if (props.isHistorical) {
+        console.log('updating partial HC results');
```
kaituo: Could you provide context of this commit?

ohltyler: Bug was found where historical HC results wasn't showing in the charts if the browser was left on the page without refreshing. This fixes that by updating the chart data whenever more progress is reported from the detector via the detector's `taskProgress` field.

public/pages/utils/anomalyResultUtils.ts
```diff
@@ -389,6 +387,8 @@ export const getAnomalySummaryQuery = (
     },
   };
 
+  // If querying RT results: remove any results that include a task_id, as this indicates
```
kaituo, line 390: Please also give some context of this commit.

ohltyler: So before this, many of the query-building fns were using `getResultFilters()` to generate the entity filters (either by model id for multi-category, or by entity.0.value for single-category). The problem was that `getResultFilters` was used in functions shared by single-entity detectors, so HC-related filters were inadvertently being added to the queries for these detectors, which was returning empty results.

This commit changes the code such that the queries will only add these filters if the detector is at least a HC detector (has an entity list length > 0). It also renames the helper fn to make it more clear that this is about entity-related filters.

kaituo:
All of the changes look good to me.  Wait for you to resolve PR build or any new commits.

ohltyler:
> All of the changes look good to me. Wait for you to resolve the PR build failure or add any new commit.

Seems identical issue with the CI is happening in main now - likely due to the version bumps in AD backend plugin, and not related to the changes in this PR. Will work on a separate fix for that.

ohltyler:
@kaituo I've re-ran again and checks are passing now. No changes were made that I know of; could be flaky tests due to low resources in GitHub runners or changes in dependent packages.

public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
```diff
@@ -266,6 +271,12 @@ export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
                         onDisplayOptionChanged={props.onDisplayOptionChanged}
                         heatmapDisplayOption={props.heatmapDisplayOption}
                         isNotSample={props.isNotSample}
+                        // Category fields in HC results are always sorted alphabetically. To make all chart
```
ylwu-amzn, line 274: Do we reply on the order of category fields to show HC detector anomaly result?

ohltyler: Yes, in the y-axis labels on the heatmap chart. From @kaituo, the results are sorted alphabetically (also see [here](https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/model/Entity.java#L126)). To be consistent on the frontend, we sort the same way.

As an example, given some detector with category fields `[a, b]`, the heatmap results will be shown with labels `<a-value>, <b-value>`. In the chart header, we list out the category field names via the header `Sort by: a, b`. If we don't sort the category field, then it is possible the listed fields would be in a different order than the shown results. For example, if the user saves a detector with category fields `[b, a]`, then without sorting, the chart header would be `Sort by: b, a`, but the results would still be `a, b`, which could lead to confusion.

public/pages/AnomalyCharts/containers/AnomalyHeatmapChart.tsx
```diff
@@ -215,9 +229,25 @@ export const AnomalyHeatmapChart = React.memo(
       return false;
     };
 
+    // Custom hook to refresh all of the heatmap data when running a historical task
```
ylwu-amzn, line 232: Will it cause heavy load if query too much? Maybe we can increase interval between refresh query for HC?

public/pages/AnomalyCharts/utils/anomalyChartUtils.tsx
```diff
@@ -252,11 +259,12 @@ export const getAnomaliesHeatmapData = (
     dateRange
   );
 
-  entityAnomaliesMap.forEach((entityAnomalies, entity) => {
+  entityAnomaliesMap.forEach((entityAnomalies, entityListAsString) => {
     const maxAnomalyGradesForEntity = [] as number[];
     const numAnomalyGradesForEntity = [] as number[];
 
-    entityValues.push(entity);
+    entityStrings.push(entityListAsString);
```
ylwu-amzn, line 266: How about add some comments for `entityStrings` and `entityLists` ?

public/pages/AnomalyCharts/containers/FeatureBreakDown.tsx
```diff
@@ -80,10 +83,19 @@ export const FeatureBreakDown = React.memo((props: FeatureBreakDownProps) => {
         const filteredFeatureData = [];
         for (let i = 0; i < anomaliesFound.length; i++) {
           const currentAnomalyData = anomaliesResult.anomalies[i];
+          const dataEntityList = get(
```
ylwu-amzn, line 86: What's the difference between `dataEntityList` and `cellEntityList`?

public/pages/AnomalyCharts/containers/AnomalyHeatmapChart.tsx
ohltyler: I think it's a fair point. I'm ok to increase the timeout - how about 30s?

public/pages/AnomalyCharts/utils/anomalyChartUtils.tsx
ohltyler: Sure, will add.

public/pages/AnomalyCharts/containers/FeatureBreakDown.tsx
ohltyler: `dataEntityList` is the associated entity list with the current anomaly data in the `for` loop. `cellEntityList` is the entity list pulled from the selected heatmap cell. Note that no logic changes were made here. Instead, `entityList` used to be `get(currentAnomalyData, 'entity', [] as EntityData[])[0].value` and `cellEntityList` used to be `props.selectedHeatmapCell.entityValue`.

kaituo:
> @kaituo I've re-ran again and checks are passing now. No changes were made that I know of; could be flaky tests due to low resources in GitHub runners or changes in dependent packages.

@ohltyler Sounds good.  I saw you will make some changes for Yaliang's comments.  Will wait for the new commit.

public/pages/AnomalyCharts/containers/AnomalyHeatmapChart.tsx
ylwu-amzn: Sounds good, maybe confirm with UX design to check if this is reasonable and should we add some tooltip

public/pages/AnomalyCharts/containers/FeatureBreakDown.tsx
ylwu-amzn: > cellEntityList is the entity list pulled from the selected heatmap cell

Does that mean user can choose multiple heatmap cells?

ohltyler: No, only one can be selected at a time. But the cell itself can contain a list of entities if it is a multi-category detector

public/pages/AnomalyCharts/containers/AnomalyHeatmapChart.tsx
ohltyler: Sure, confirming with UX.

ohltyler: Confirmed 30s is ok - no extra notices/wording is necessary.

kaituo (approved):

ylwu-amzn (approved):

PR #69, opened 2021-08-17
Fix BWC for legacy detectors

ohltyler:
Signed-off-by: Tyler Ohlsen ohltyler@amazon.com

### Description

This PR fixes backward compatibility for realtime and historical detectors created on versions 1.0 or before.

The changes allow for the following use cases to occur:
1. User creates realtime detector on 1.0 => upgrades to 1.1 => renders as a detector running a realtime job, with no associated historical analysis
2. User creates historical detector on 1.0 => upgrades to 1.1 => renders as a detector with no realtime job, and a populated historical analysis

For 1.1, the backend detector data models & task data models have changed, but we maintain the same data models on the frontend. To handle this, the frontend needs to be able to properly parse data coming from either of the data models (either 1.1 or pre-1.1).

The solution consists of having separate logic for parsing 2 different kinds of detector data:
1. Static info (detector ID / name / description / timestamp, etc). This is data that has remained in the same format across the different data models. It does not dynamically change based on a user starting or stopping a realtime or historical task. This type of data can be pulled and parsed in similar ways as before.
2. Task-related & job-related info (curState / initProgress / taskId / taskProgress, etc.). Prior to 1.1, this data was pulled from different APIs (get detector API, profile API), and was parsed according to the formatted data that was returned. Because the returned format has changed, trying to parse this data from these responses can get extremely complex when considering all of the different scenarios, especially if Dashboards is communicating with a mixed cluster (e.g., old detector on old node, new detector on old node, old detector on new node, new detector on new node). To simplify where this data comes from, we can instead use the search task API to pull all of the latest realtime & historical tasks per detector. This decouples the implementation-specific formats with the task-and-job-related detector info.

One way that this search task strategy as described in point 2 fails, is when trying to fetch info for old realtime detectors. These types of detectors don't actually have any existing tasks. For consistency, the backend will perform a one-time backfilling task, which will create realtime tasks for any old realtime detector, once a node running on the new 1.1 version has joined a cluster. Before the backfilling is complete, however, the frontend has no way of knowing an old realtime detector's true state, without making excess API calls. To handle this, the frontend will estimate such detectors as `Running` or `Stopped`, based on any associated job set to `enabled` or not, respectively. If no job is found, the state is defaulted to `Stopped`.

Some of the other ways BWC is supported here includes:
- searching over the legacy `HISTORICAL` task type to fetch legacy historical tasks
- handling the possibility of multiple realtime tasks returned, and selecting the most recent one based on the `execution_start_time` (this case can possibly happen when performing the legacy realtime task backfilling)
- pulling `detection_date_range` from 2 different possible places, since the field was refactored in the 1.1 changes

Testing done:
- Created detectors on 1.0 => upgraded domain to 1.1-target-changes => make sure render properly. Specifically, the following scenarios were tested:
     - Realtime detector (never ran)
     - Realtime detector (running)
     - Realtime high-cardinality detector (never ran)
     - Realtime high-cardinality detector (running)
     - Historical detector (never ran)
     - Historical detector (ran)
     - Detector with no features
     - Sample detector
- After upgrading, each was tested by running an additional realtime job and historical analysis to make sure the stored tasks were updated to show the latest.
- New detectors were created on the new domain for all of the same detector types as listed above (except for the no-features-detector). Additional realtime jobs and historical analyses were ran for them as well, to make sure the stored tasks were updated to show the latest.
- Added UT for the helper fns added in `adHelpers`

### Check List


server/routes/utils/adHelpers.ts
```diff
@@ -436,3 +526,59 @@ export const getLatestDetectorTasksQuery = () => {
...
+          {
+            term: {
+              is_latest: 'true',
+            },
+          },
+          {
+            terms: {
+              task_type: [
+                'REALTIME_HC_DETECTOR',
+                'REALTIME_SINGLE_ENTITY',
```
ylwu-amzn, line 550: Plan to change `REALTIME_SINGLE_ENTITY` to `REALTIME_SINGLE_FLOW` to avoid confusion with HC entity.

server/routes/utils/adHelpers.ts
```diff
@@ -436,3 +526,59 @@ export const getLatestDetectorTasksQuery = () => {
...
+
+// Helper fn to get the latest realtime task for some detector.
+// It is possible that the backend returns multiple realtime tasks
+// set to 'is_latest=true' when backfilling legacy realtime detectors.
+export const getLatestRealtimeTask = (taskList: any[]) => {
+  const realtimeTaskList = taskList.filter((task: any) => isRealTimeTask(task));
+  let realtimeTask = get(realtimeTaskList, 0);
+  if (get(realtimeTaskList, 'length', 0) > 1) {
+    realtimeTask =
+      get(realtimeTaskList, '0.execution_start_time') >
```
ylwu-amzn, line 571: Can we just sort by `execution_start_time` and get the first realtime task?

server/routes/utils/adHelpers.ts
ohltyler: I know backend already uses 'single entity' in many places to describe these kinds of detectors. Can you explain the confusion?

server/routes/utils/adHelpers.ts
ohltyler: Since the detectors can return 3 possible tasks (1 historical, 2 realtime), it is possible that a historical task is lost if the 2 realtime tasks both have newer `execution_start_time`s. We could sort if we did separate queries for realtime and historical, but prefer to keep as one. Let me know what you think.

server/routes/utils/adHelpers.ts
ylwu-amzn: I remember Kaituo mentioned some word changes, can't recall the details. @kaituo do you have any concern about the single entity name ? If it's ok, I'm fine to keep `_SINGLE_ENTITY`

server/routes/utils/adHelpers.ts
ohltyler: Recently learned that during the backfilling process on backend, that multiple realtime tasks can be created during some edge cases as new nodes join the cluster (multiple nodes join at the same time and see that no realtime tasks exist for some detector, so both create them). And even more recently learned that it may not be just 2, but an arbitrary amount, if x number of nodes join the cluster and trigger this edge case.

Because of this, I'm thinking of changing to Yaliang's suggestion of having a separate query to fetch realtime tasks, sort by `execution_start_time`, and return the top 1. But not sure if the performance of making 2 search task API calls vs. 1 call is worth the performance hit, since it is just for handling this edge case.

server/routes/ad.ts
```diff
@@ -245,30 +249,54 @@ export default class AdService {
...
-        stateError: processTaskError(
-          get(response, 'realtime_detection_task.error', '')
-        ),
-        initProgress: getTaskInitProgress(response.realtime_detection_task),
+      // Populating static detector fields
+      const staticFields = {
+        id: detectorResponse._id,
+        primaryTerm: detectorResponse._primary_term,
+        seqNo: detectorResponse._seq_no,
+        ...convertStaticFieldsToCamelCase(detectorResponse.anomaly_detector),
```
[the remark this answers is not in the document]
ohltyler: Frontend converts to camelCase to follow standard javascript naming conventions.

public/pages/DetectorDetail/containers/DetectorDetail.tsx
```diff
@@ -226,6 +227,7 @@ export const DetectorDetail = (props: DetectorDetailProps) => {
   const handleStartAdJob = async (detectorId: string) => {
     try {
       await dispatch(startDetector(detectorId));
+      dispatch(getDetector(detectorId));
```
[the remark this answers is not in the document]
ohltyler: `await` is used to wait for the promise returned after calling `startDetector`, to make sure the detector was started correctly, so we can show the success toast. We don't care about waiting for the `getDetector` call here, and instead will just make the call, and update the page once the call returns at some later time.

The reason for adding this line here at all, is because after starting the detector, we need to re-fetch the latest RT and historical tasks to show the latest info.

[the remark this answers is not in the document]
kaituo: Could you add these to the comments?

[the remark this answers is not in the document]
ohltyler: Sure, will add.

server/routes/utils/adHelpers.ts
ylwu-amzn: I'm not sure about the how likely the edge case will happen, but it will happen if some new node starts first, then it start to backfill ,then other nodes join cluster, the AD version hash ring may change, so other node will backfill again. It will be safe to query separately for realtime and historical.

ohltyler: Got it. Will change to have separate queries.

ohltyler: @kaituo @ylwu-amzn changed to use separate queries in latest commit.

server/routes/utils/adHelpers.ts
```diff
@@ -532,9 +532,16 @@ export const isRealTimeTask = (task: any) => {
...
+export const getLatestTaskForDetectorQuery = (
+  detectorId: string,
+  realtime: boolean
+) => {
+  const taskTypes = realtime ? REALTIME_TASK_TYPES : HISTORICAL_TASK_TYPES;
   return {
-    size: 3,
+    size: 1,
+    sort: {
+      execution_start_time: SORT_DIRECTION.DESC,
```
kaituo, line 543: Could you add your explanation why there can be multiple real time tasks and we need to sort as comments? https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/69#discussion_r694227657

ohltyler: Sure, will add.

ohltyler: Added in this [commit](https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/69/commits/fa3b25eb3e9f8aca606fba02d191f9d5b609df28)

kaituo (approved):

ylwu-amzn (approved):
LGTM. Thanks for the change!

PR #78, opened 2021-08-30
Enable zooming in HC entity charts

### LINKED ISSUE #75: Support zooming on selected heatmap cell anomaly charts
Currently, we support zooming and daily/weekly/monthly aggregations for historical detectors, but only for single-entity detectors. For high-cardinality detectors, when a user selects a heatmap cell, the displayed chart doesn't support any zooming or aggregation functionality. Even at only 1 year of historical data, the charts can get difficult to read, where any sort of zooming capability would be helpful for the user (see screenshot below).

Suggest to support the same zooming (and maybe daily/weekly/monthly aggregations) for the displayed anomaly charts when the user selects a heatmap cell. This adds consistency across the detector types and easier readability for high-cardinality historical detectors.

Screen Shot 2021-08-25 at 9 13 49 PM

ohltyler:

### Description

This PR adds zooming capabilities to the anomaly chart when selecting a heatmap cell for high-cardinality detectors.

Changes include:
- `AnomaliesDetailChart`: Allowing zooming for high-cardinality detectors
- `FeatureBreakDown`: Setting the date range based on the passed-in date range (which is `zoomRange`) instead of the heatmap cell. This is so we can narrow the range if the user zooms, rather than leaving it at the heatmap cell's static range.
- `AnomalyHistory`: setting the `zoomRange` when a heatmap cell is selected or de-selected, to refresh any stored data in the components. This is needed since `FeatureBreakDown` now relies on the `zoomRange` to get the date range. 
- `AnomalyResultsTable`: Only populate with anomalies when applicable. The two applicable cases are (1) non high-cardinality detectors and (2) high-cardinality detectors when a heatmap cell is selected. This had to be changed due to the `zoomRange` allowing more results to be shown, so more filtering was needed.

Testing:
Confirmed the zooming works, and updates the corresponding table and feature breakdown accordingly.
Confirmed that selecting different heatmap cells or de-selecting heatmap cells will clear the zoom range.
Confirmed it works for historical and real-time.
Confirmed it doesn't break the components for non-high-cardinality detectors.

### Issues Resolved

#75

### Check List

### DIFF
--- a/public/pages/AnomalyCharts/containers/AnomalyDetailsChart.tsx
+++ b/public/pages/AnomalyCharts/containers/AnomalyDetailsChart.tsx
@@ -494,28 +494,24 @@ export const AnomalyDetailsChart = React.memo(
                     //TODO: research more why only set this old property will work.
                     showLegendDisplayValue={false}
                     legendPosition={Position.Right}
-                    onBrushEnd={
-                      props.isHCDetector
-                        ? undefined
-                        : (brushArea: XYBrushArea) => {
-                            const start = get(
-                              brushArea,
-                              'x.0',
-                              DEFAULT_DATE_PICKER_RANGE.start
-                            );
-                            const end = get(
-                              brushArea,
-                              'x.1',
-                              DEFAULT_DATE_PICKER_RANGE.start
-                            );
-                            !props.bucketizedAnomalies
-                              ? handleZoomRangeChange(start, end)
-                              : handleDateRangeChange(start, end);
-                            if (props.onDatePickerRangeChange) {
-                              props.onDatePickerRangeChange(start, end);
-                            }
-                          }
-                    }
+                    onBrushEnd={(brushArea: XYBrushArea) => {
+                      const start = get(
+                        brushArea,
+                        'x.0',
+                        DEFAULT_DATE_PICKER_RANGE.start
+                      );
+                      const end = get(
+                        brushArea,
+                        'x.1',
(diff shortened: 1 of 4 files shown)


public/pages/AnomalyCharts/containers/AnomalyDetailsChart.tsx
```diff
@@ -494,28 +494,24 @@ export const AnomalyDetailsChart = React.memo(
...
-                            );
-                            !props.bucketizedAnomalies
-                              ? handleZoomRangeChange(start, end)
-                              : handleDateRangeChange(start, end);
-                            if (props.onDatePickerRangeChange) {
-                              props.onDatePickerRangeChange(start, end);
-                            }
-                          }
-                    }
+                    onBrushEnd={(brushArea: XYBrushArea) => {
```
ylwu-amzn, line 497: For single-entity detector, the zooming will dynamically sample anomaly results and show. For example, if the date range is big and contains too many data points, we will sample 1000 anomaly results points and show them. For HC detector, is it the same logic?

ohltyler: Yes, we bucketize the results the same way as the single entity. It is abstracted such that the results are handled without context of single entity vs. multi entity, real-time vs. historical, etc.

ylwu-amzn (approved):

kaituo (approved):

PR #107, opened 2021-10-28
Support multi-category filtering of results

ohltyler:
### Description

This PR adds support for filtering results on a subset of categorical fields. For example, if a user has 2 categorical fields selected, they can now search for the top values of one field, ignoring the other. And, when selecting a value of one of these fields, a sorted list of the top values of the other field is displayed for viewing on the anomaly and feature charts. This is supported for both real-time and historical scenarios.

The primary changes in this PR include:
- Adding support for new multi-category filtering API (see [backend PR](https://github.com/opensearch-project/anomaly-detection/pull/261))
- Adding a combo box filter at the top of the charts to select some subset of categorical fields, rather than a hardcoded string that includes all of the fields.
- Adding a combo box filter when a heatmap cell is selected, to select multiple values to display on the charts.
- Adding support for multiple time series on the anomaly charts and feature charts.
- Moving anomaly filtering / sorting logic from the child components into the base container (`AnomalyHistory`), making them more stateless as components should be.
- Minor changes to the entity list format in the heatmap cell. Changed from `Entities: <entity-value-1>, <entity-value-2>` => `Entities: <entity-name-1>: <entity-value1>, <entity-name-2>: <entity-value-2>`. The main reason for this change is so the full entity information can be passed to the child components when needed - before, only the values were being passed.
- Fixing bucketized results - previous code had a bug where bucketized results weren't being retrieved properly. That has been fixed here, which also meant more logic needed to be added to handle zooming from bucketized results => non-bucketized results.
- Misc. other cleanup (fixed stale data occasionally showing up on feature and details charts, fixed sample anomaly zoom state bug, separated logic of sample vs. non-sample data in heatmap chart, added comments for existing vars or methods that weren't clear, renamed many vars and methods to make them more clear)

Testing:
- All UT pass
- All IT pass
- Confirmed it is consistent for real-time and historical
- Confirmed nothing breaks on non-high-cardinality & high-cardinality-single-category detector results
- Confirmed nothing breaks on sample anomalies via preview

Screenshots:

New combo box filter for categorical fields:
Screen Shot 2021-10-28 at 12 57 11 PM

New combo box filter for child categorical field values (when a heatmap cell is selected):
Screen Shot 2021-10-28 at 12 57 42 PM

Multi-time-series support, with consistent coloring for anomaly & feature results:
Screen Shot 2021-10-28 at 12 58 08 PM

### Check List

### DIFF
--- a/public/pages/AnomalyCharts/components/FeatureChart/FeatureChart.tsx
+++ b/public/pages/AnomalyCharts/components/FeatureChart/FeatureChart.tsx
@@ -83,6 +83,7 @@ interface FeatureChartProps {
   detectorEnabledTime?: number;
   rawFeatureData: FeatureAggregationData[][];
   entityData: EntityData[][];
+  isHCDetector?: boolean;
 }
 
 export const FeatureChart = (props: FeatureChartProps) => {
@@ -174,6 +175,8 @@ export const FeatureChart = (props: FeatureChartProps) => {
     props.dateRange
   ) as FeatureAggregationData[][];
 
+  const multipleTimeSeries = get(featureData, 'length', 0) > 1;
+
   return (
     <ContentPanel
       title={
@@ -200,7 +203,7 @@ export const FeatureChart = (props: FeatureChartProps) => {
           opacity: showLoader ? 0.2 : 1,
         }}
       >
-        <Chart>
+        <Chart key={`${featureData}`}>
           <Settings
             showLegend
             showLegendExtra={false}
@@ -258,9 +261,11 @@ export const FeatureChart = (props: FeatureChartProps) => {
           }
           {featureData.map(
             (featureTimeSeries: FeatureAggregationData[], index) => {
-              const multipleTimeSeries = get(props, 'entityData.length', 0) > 1;
-              const seriesKey = multipleTimeSeries
-                ? convertToEntityString(props.entityData[index], ', ')
+              const seriesKey = props.isHCDetector
+                ? `${props.featureDataSeriesName} (${convertToEntityString(
+                    props.entityData[index],
+                    ', '
+                  )})`
                 : props.featureDataSeriesName;
               return (
                 <LineSeries
--- a/public/pages/AnomalyCharts/components/FeatureChart/FeatureChart.tsx
+++ b/public/pages/AnomalyCharts/components/FeatureChart/FeatureChart.tsx
@@ -203,6 +203,11 @@ export const FeatureChart = (props: FeatureChartProps) => {
           opacity: showLoader ? 0.2 : 1,
         }}
       >
+        {/**
+         * Charts may show stale data even after feature data is changed.
+         * By setting the key prop here, the chart will re-mount whenever the
+         * feature data has changed.
+         */}
         <Chart key={`${featureData}`}>
           <Settings
             showLegend
--- a/public/pages/AnomalyCharts/components/FeatureChart/FeatureChart.tsx
+++ b/public/pages/AnomalyCharts/components/FeatureChart/FeatureChart.tsx
@@ -46,20 +46,28 @@ import {
   DateRange,
   FEATURE_TYPE,
   Schedule,
+  EntityData,
 } from '../../../../models/interfaces';
 import { darkModeEnabled } from '../../../../utils/opensearchDashboardsUtils';
 import {
   prepareDataForChart,
   getFeatureMissingDataAnnotations,
   filterWithDateRange,
+  flattenData,
+  convertToEntityString,
 } from '../../../utils/anomalyResultUtils';
 import { CodeModal } from '../../../DetectorConfig/components/CodeModal/CodeModal';
-import { CHART_FIELDS, FEATURE_CHART_THEME } from '../../utils/constants';
-import { isEmpty } from 'lodash';
+import {
+  CHART_FIELDS,
+  CHART_COLORS,
+  FEATURE_CHART_THEME,
+} from '../../utils/constants';
+import { get, isEmpty } from 'lodash';
+import { ENTITY_COLORS } from '../../../DetectorResults/utils/constants';
 
 interface FeatureChartProps {
   feature: FeatureAttributes;
-  featureData: FeatureAggregationData[];
+  featureData: FeatureAggregationData[][];
   annotations: any[];
   isLoading: boolean;
   dateRange: DateRange;
@@ -73,7 +81,8 @@ interface FeatureChartProps {
   detectorInterval: Schedule;
   showFeatureMissingDataPointAnnotation?: boolean;
   detectorEnabledTime?: number;
-  rawFeatureData: FeatureAggregationData[];
+  rawFeatureData: FeatureAggregationData[][];
+  entityData: EntityData[][];
 }
 
 export const FeatureChart = (props: FeatureChartProps) => {
@@ -128,8 +137,6 @@ export const FeatureChart = (props: FeatureChartProps) => {
     </EuiText>
   );
 
-  const featureData = prepareDataForChart(props.featureData, props.dateRange);
-
   // return undefined if featureMissingDataPointAnnotationStartDate is missing
   // OR it is even behind the specified date range
   const getFeatureMissingAnnotationDateRange = (
@@ -161,6 +168,12 @@ export const FeatureChart = (props: FeatureChartProps) => {
       'coordinates.x0'
     );
   };
+
+  const featureData = prepareDataForChart(
+    props.featureData,
+    props.dateRange
+  ) as FeatureAggregationData[][];
+
   return (
     <ContentPanel
       title={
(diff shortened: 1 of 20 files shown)


public/pages/AnomalyCharts/components/FeatureChart/FeatureChart.tsx
```diff
@@ -187,7 +203,12 @@ export const FeatureChart = (props: FeatureChartProps) => {
           opacity: showLoader ? 0.2 : 1,
         }}
       >
-        <Chart>
+        {/**
+         * Charts may show stale data even after feature data is changed.
```
ylwu-amzn, line 207: You mean even `props.featureData` changed the chart won't refresh itself? Looks like some bug of `Chart`?

public/pages/utils/anomalyResultUtils.ts
```diff
@@ -56,13 +60,14 @@ import {
...
   FeatureAttributes,
+  MonitorAlert,
 } from '../../models/interfaces';
 import { getDetectorLiveResults } from '../../redux/reducers/liveAnomalyResults';
 import {
   MAX_ANOMALIES,
   MISSING_FEATURE_DATA_SEVERITY,
 } from '../../utils/constants';
-import { HeatmapCell } from '../AnomalyCharts/containers/AnomalyHeatmapChart';
+//import { HeatmapCell } from '../AnomalyCharts/containers/AnomalyHeatmapChart';
```
ylwu-amzn, line 70: minor: remove this line

public/pages/utils/anomalyResultUtils.ts
```diff
@@ -231,72 +237,135 @@ export const prepareDataForLiveChart = (
...
-    confidence: null,
-    anomalyGrade: null,
+  return preparedData;
+};
+
+export const generateAnomalyAnnotations = (
+  anomalies: AnomalyData[][]
+): any[][] => {
+  let annotations = [] as any[];
+  anomalies.forEach((anomalies: AnomalyData[]) => {
```
ylwu-amzn, line 264: Looks confusing as we have two `anomalies` inside and outside of `forEach` . How about rename the `anomalies` inside `forEach` ?

public/pages/AnomalyCharts/utils/anomalyChartUtils.tsx
```diff
@@ -190,8 +203,8 @@ export const ANOMALY_HEATMAP_COLORSCALE = [
 ];
 
 export enum AnomalyHeatmapSortType {
-  SEVERITY = 'by_severity',
-  OCCURRENCES = 'by_occurrences',
+  SEVERITY = 'severity',
+  OCCURRENCES = 'occurrence',
```
amitgalitz, line 207: minor but should maybe change this from OCCURRENCES to OCCURRENCE?

public/pages/utils/anomalyResultUtils.ts
```diff
@@ -231,72 +237,135 @@ export const prepareDataForLiveChart = (
...
-    startTime: dateRange.endDate,
-    endTime: dateRange.endDate,
-    plotTime: dateRange.endDate,
-    confidence: null,
-    anomalyGrade: null,
+  return preparedData;
+};
+
+export const generateAnomalyAnnotations = (
+  anomalies: AnomalyData[][]
```
ylwu-amzn, line 261: How about add some comments about what each dimension means for this 2D array `AnomalyData`

public/pages/AnomalyCharts/components/FeatureChart/FeatureChart.tsx
ohltyler: Tried debugging for awhile, could either be a bug in EUI's `Chart` or because of the fact that `featureData` is an array, and there's some caveats where arrays as state variables won't trigger an update, if the values are simply changed (see https://stackoverflow.com/questions/56266575/why-is-usestate-not-triggering-re-render). I've tried some of the suggestions there with no luck, so I'm guessing some issue with `Chart`.

I was able to see the list get updated and log outputs that it is updated, but the chart remains the same. This fix just guarantees a re-render.

public/pages/utils/anomalyResultUtils.ts
ohltyler: Thanks, will remove

public/pages/utils/anomalyResultUtils.ts
ohltyler: Good catch, will rename.

public/pages/AnomalyCharts/utils/anomalyChartUtils.tsx
ohltyler: Fair point, I'll change.

public/pages/utils/anomalyResultUtils.ts
ohltyler: Sure, will add.

public/pages/AnomalyCharts/components/FeatureChart/FeatureChart.tsx
ylwu-amzn: Got it. If set the key as `featureData`, how will the rendered HTML look like? The id will be the the whole content of `featureData` ? I'm ok with this workaround, just curious if it will generate long id.

ohltyler: Interesting question - I checked and don't see it getting rendered in HTML at all. I believe that's because `key` is a react-specific prop for allowing user-controlled re-rendering, and isn't necessary to include in the produced HTML. `id` is an HTML-specific attribute, which would get rendered in HTML.

public/pages/AnomalyCharts/components/FeatureChart/FeatureChart.tsx
```diff
@@ -240,15 +261,35 @@ export const FeatureChart = (props: FeatureChartProps) => {
...
-            xAccessor={CHART_FIELDS.PLOT_TIME}
-            yAccessors={[CHART_FIELDS.DATA]}
-            data={featureData}
-          />
+          {
+            // Add each set of feature data as a separate time series
+          }
+          {featureData.map(
+            (featureTimeSeries: FeatureAggregationData[], index) => {
+              const seriesKey = props.isHCDetector
```
amitgalitz, line 269: Might be misunderstanding typescript 'types' but since `isHCDetector` is an optional, means `props.isHCDetector` is a 'falsy' value and lead to the case after `:`. Is this on purpose or could it lead to any issues.

public/pages/AnomalyCharts/components/FeatureChart/FeatureChart.tsx
```diff
@@ -217,8 +238,8 @@ export const FeatureChart = (props: FeatureChartProps) => {
                   domainType={AnnotationDomainTypes.XDomain}
                   dataValues={getFeatureMissingDataAnnotations(
```
amitgalitz, line 239: nit: Confused what the featureMissingDataAnnotations are (could just need an explanation offline).

public/pages/AnomalyCharts/containers/AnomalyDetailsChart.tsx
```diff
@@ -137,7 +136,9 @@ export const AnomalyDetailsChart = React.memo(
     const [zoomRange, setZoomRange] = useState<DateRange>({
       ...props.dateRange,
     });
-    const [zoomedAnomalies, setZoomedAnomalies] = useState<any[]>([]);
+    const [zoomedAnomalies, setZoomedAnomalies] = useState<AnomalyData[][]>(
+      props.anomalies
+    );
 
     const [aggregatedAnomalies, setAggregatedAnomalies] = useState<any[]>([]);
```
amitgalitz, line 143: nit: seems like this can be changed from `useState<any[]>` to `useState<anomalyData[]>`. Seems to only be set as that type, not sure if changing will cause further issues down the line when using the variable.

public/pages/AnomalyCharts/components/FeatureChart/FeatureChart.tsx
ohltyler: It's on purpose - if `isHCDetector` exists and is `true`, then the true condition runs; else the false condition runs.

public/pages/AnomalyCharts/components/FeatureChart/FeatureChart.tsx
ohltyler: It's annotations on the chart to show if a user has sparse / missing data on the feature chart, where feature results couldn't be generated due to no data (e.g., no data in 10-min period w/ an interval of 10 mins may lead to no features, therefore no anomaly results).

public/pages/AnomalyCharts/containers/AnomalyDetailsChart.tsx
ohltyler: I had changed it from `any[]` to `AnomalyData[][]`, it needs to be a 2D arr now. Not sure what you mean, any other concerns?

amitgalitz: My mistake I mean to highlight line 143 `const [aggregatedAnomalies, setAggregatedAnomalies] = useState<any[]>([]);` but also it's not something you seem to change this PR so probably unnecessary for now.

ohltyler: Good catch actually, added type safety, verified results filtering/aggregation working as expected.

public/pages/AnomalyCharts/components/FeatureChart/FeatureChart.tsx
ylwu-amzn: Got it, that should be ok

ylwu-amzn (approved):

amitgalitz (dismissed):

ylwu-amzn (approved):

PR #116, opened 2021-11-05
Fix e2e

amitgalitz:
Signed-off-by: Amit Galitzky amgalitz@amazon.com

### Description

Fixed E2E workflow so it doesn't break everytime since it used to run based of a docker image which we don't use anymore. Instead of using docker I run Opensearch with AD plugin using the 1.2.0 Snapshot. Currently I didn't distinguish two different versions like we used to have with and without security.

### Check List

### DIFF
--- a/.github/workflows/e2e-tests-workflow.yml
+++ b/.github/workflows/e2e-tests-workflow.yml
@@ -27,7 +27,7 @@ jobs:
           ref: 'main'
       - name: Run Opensearch with plugin
         run: |
-          cd anoamly-detection
+          cd anomaly-detection
             ./gradlew run -Dopensearch.version=${{ env.OPENSEARCH_VERSION }} &
             timeout 300 bash -c 'while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' localhost:9200)" != "200" ]]; do sleep 5; done'
       - name: Checkout OpenSearch Dashboards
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e-tests-workflow.yml
```diff
@@ -8,10 +8,7 @@ on:
...
 env:
   OPENSEARCH_DASHBOARDS_VERSION: '1.x'
-  AD_OPENSEARCH_DASHBOARDS_PLUGIN_NAME: anomaly-detection-dashboards
-  OPENSEARCH_DOCKER_IMAGE: opensearchstaging/opensearch
-  DASHBOARDS_DOCKER_IMAGE: opensearchstaging/opensearch-dashboards
-  DOCKER_TAG: 1.1.0
+  OPENSEARCH_VERSION: '1.2.0-SNAPSHOT'
 jobs:
   test-with-security:
     name: Run e2e tests with security
```
ohltyler, line 14: the regression here is now we're only testing with no security (this can be renamed). Part of the reason for using the docker images was that security was included. Is there a way to pull security dependencies from maven?

amitgalitz: Looking into it

ohltyler (dismissed):
Approving to unblock. We can fix security-enabled tests in CI after if necessary.

ylwu-amzn (approved):

PR #122, opened 2021-11-08
Fix chart range bug of non-HC anomaly and feature charts

### LINKED ISSUE #121: [BUG] Anomaly overview chart shows wrong range on time axis
Start a new single-entity detector with 1 minute interval, wait for 3 minutes, then check realtime results page for last 24 hours. The time axis range doesn't show last 24 hours, it just shows last 3 minutes
Screen Shot 2021-11-07 at 8 18 44 PM

ohltyler:

### Description

Fixes a bug in the non-high-cardinality scenario where anomaly and feature charts were showing the explicit range based on the timestamps of anomaly results found in the parent date range, rather than showing the static date range itself. (see #121 for details/screenshots).

This fixes that bug by 2 main changes:
1. Explicitly setting the range of `x` values for the chart via `xDomain` chart prop for both anomaly details and feature charts
2. Splitting up the logic for HC and non-HC results, when the zoom range is changed. For example, in the non-HC scenario, if the zoom range is changed, we want to update the parent date range (as shown in the date range picker). In the HC scenario, we only want to change the range shown on the child charts, and not on the parent date range (if we change the parent date range here, then any selected heatmap cell would be de-selected, and the entire heatmap would be re-generated based on the new date range, which we don't want).

Confirmed everything works as expected for non-HC, single-HC, and multi-HC, for both RT and historical results.

### Issues Resolved

Closes #121 

### Check List

### DIFF
--- a/public/pages/AnomalyCharts/containers/AnomalyDetailsChart.tsx
+++ b/public/pages/AnomalyCharts/containers/AnomalyDetailsChart.tsx
@@ -304,11 +304,18 @@ export const AnomalyDetailsChart = React.memo(
     }, [props.anomalies, zoomRange, aggregatedAnomalies, selectedAggId]);
 
     const handleZoomRangeChange = (start: number, end: number) => {
-      setZoomRange({
-        startDate: start,
-        endDate: end,
-      });
-      props.onZoomRangeChange(start, end);
+      // In the HC scenario, we only want to change the local zoom range.
+      // We don't want to change the overall date range, since that would auto-de-select
+      // any selected heatmap cell, and re-fetch results based on the new date range
+      if (props.isHCDetector) {
+        setZoomRange({
+          startDate: start,
+          endDate: end,
+        });
+        props.onZoomRangeChange(start, end);
+      } else {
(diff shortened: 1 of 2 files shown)


public/pages/AnomalyCharts/containers/AnomalyDetailsChart.tsx
```diff
@@ -505,14 +512,22 @@ export const AnomalyDetailsChart = React.memo(
...
                       );
                       handleZoomRangeChange(start, end);
                       if (props.onDatePickerRangeChange) {
                         props.onDatePickerRangeChange(start, end);
                       }
                     }}
                     theme={ANOMALY_CHART_THEME}
+                    xDomain={
+                      showAggregateResults
+                        ? undefined
```
ohltyler, line 525: an error will be thrown if we set an `xDomain` in the aggregate anomalies case, since the scale type (see lines 633-635) won't work properly in the `ScaleType.Ordinal` type

ylwu-amzn (approved):

amitgalitz (approved):

PR #126, opened 2021-11-09
search only custom result index on detector detail page

ylwu-amzn:
Signed-off-by: Yaliang Wu <ylwu@amazon.com>

### Description

Only query custom result index for detector detail page. Query both default and custom AD result index on dashboard and detector list page.

### Check List

  - [ ] All tests pass

### DIFF
--- a/public/pages/AnomalyCharts/containers/AnomalyDetailsChart.tsx
+++ b/public/pages/AnomalyCharts/containers/AnomalyDetailsChart.tsx
@@ -168,7 +168,7 @@ export const AnomalyDetailsChart = React.memo(
         zoomRange.endDate,
         taskId
       );
-      dispatch(searchResults(anomalyDataRangeQuery, resultIndex))
+      dispatch(searchResults(anomalyDataRangeQuery, resultIndex, true))
         .then((response: any) => {
           // Only retrieve buckets that are in the anomaly results range. This is so
           // we don't show aggregate results for where there is no data at all
@@ -187,7 +187,7 @@ export const AnomalyDetailsChart = React.memo(
             taskId,
             selectedAggId
           );
-          dispatch(searchResults(historicalAggQuery, resultIndex))
+          dispatch(searchResults(historicalAggQuery, resultIndex, true))
             .then((response: any) => {
               const aggregatedAnomalies = parseHistoricalAggregatedAnomalies(
                 response,
@@ -223,7 +223,7 @@ export const AnomalyDetailsChart = React.memo(
           zoomRange.endDate,
           taskId
         );
-        dispatch(searchResults(anomalyDataRangeQuery, resultIndex))
+        dispatch(searchResults(anomalyDataRangeQuery, resultIndex, true))
           .then((response: any) => {
             const dataStartDate = get(
               response,
--- a/public/pages/Dashboard/Components/AnomaliesDistribution.tsx
+++ b/public/pages/Dashboard/Components/AnomaliesDistribution.tsx
@@ -70,7 +70,8 @@ export const AnomaliesDistributionChart = (
         dispatch,
         0,
         false,
-        ALL_CUSTOM_AD_RESULT_INDICES
+        ALL_CUSTOM_AD_RESULT_INDICES,
+        false
       );
     setAnomalyDistribution(distributionResult);
 
--- a/public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
+++ b/public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
@@ -101,7 +101,8 @@ export const AnomaliesLiveChart = (props: AnomaliesLiveChartProps) => {
         -1,
         1,
         true,
-        ALL_CUSTOM_AD_RESULT_INDICES
+        ALL_CUSTOM_AD_RESULT_INDICES,
+        false
       );
     } catch (err) {
       console.log(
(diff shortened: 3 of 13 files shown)


ohltyler:
So looks like this adds a param to the following redux functions:
`getDetectorResults()`
`searchResults()`
`getDetectorLiveResults()`

which ends up making calls to the following server-side functions:
`searchResults()`
`getDetectors()` (which sets the new param as `false` to guarantee results for all detectors)
`getAnomalyResults()`

Any reason why the logic for determining whether `onlyQueryCustomResultIndex` should be `true/false` can't be determined within the server-side functions alone? From my understanding, on client-side, passing `resultIndex = ALL_CUSTOM_AD_RESULT_INDICES` always corresponds to `onlyQueryCustomResultIndex = false`, and vice versa. So couldn't the server-side functions determine `onlyQueryCustomResultIndex` by checking the value of `resultIndex.startsWith(CUSTOM_AD_RESULT_INDEX_PREFIX)`? Seems this could minimize the overall code changes, and not add any extra params which have to be passed from client-side -> server-side

amitgalitz:
Is this change only for detectors that themselves have a custom result index? Not across the board for every detector detail page right?

ylwu-amzn:
> So looks like this adds a param to the following redux functions: `getDetectorResults()` `searchResults()` `getDetectorLiveResults()`
> 
> which ends up making calls to the following server-side functions: `searchResults()` `getDetectors()` (which sets the new param as `false` to guarantee results for all detectors) `getAnomalyResults()`
> 
> Any reason why the logic for determining whether `onlyQueryCustomResultIndex` should be `true/false` can't be determined within the server-side functions alone? From my understanding, on client-side, passing `resultIndex = ALL_CUSTOM_AD_RESULT_INDICES` always corresponds to `onlyQueryCustomResultIndex = false`, and vice versa. So couldn't the server-side functions determine `onlyQueryCustomResultIndex` by checking the value of `resultIndex.startsWith(CUSTOM_AD_RESULT_INDEX_PREFIX)`? Seems this could minimize the overall code changes, and not add any extra params which have to be passed from client-side -> server-side

Yeah, we can do that. Here I expose the parameter to make it more flexible for client side. If we want to query only `ALL_CUSTOM_AD_RESULT_INDICES` in future like showing how many anomalies in custom result indices, we can just pass `true` to server-side function. And this also reminds other engineers who don't know the backend implementation details that they should consider whether just query custom result index is enough.

ylwu-amzn:
> Is this change only for detectors that themselves have a custom result index? Not across the board for every detector detail page right?

Yes, if detector is not using custom result index, we just query default result index. This change is mainly to improve performance for detector using custom result index by not querying default result index.

ohltyler:
> > So looks like this adds a param to the following redux functions: `getDetectorResults()` `searchResults()` `getDetectorLiveResults()`
> > which ends up making calls to the following server-side functions: `searchResults()` `getDetectors()` (which sets the new param as `false` to guarantee results for all detectors) `getAnomalyResults()`
> > Any reason why the logic for determining whether `onlyQueryCustomResultIndex` should be `true/false` can't be determined within the server-side functions alone? From my understanding, on client-side, passing `resultIndex = ALL_CUSTOM_AD_RESULT_INDICES` always corresponds to `onlyQueryCustomResultIndex = false`, and vice versa. So couldn't the server-side functions determine `onlyQueryCustomResultIndex` by checking the value of `resultIndex.startsWith(CUSTOM_AD_RESULT_INDEX_PREFIX)`? Seems this could minimize the overall code changes, and not add any extra params which have to be passed from client-side -> server-side
> 
> Yeah, we can do that. Here I expose the parameter to make it more flexible for client side. If we want to query only `ALL_CUSTOM_AD_RESULT_INDICES` in future like showing how many anomalies in custom result indices, we can just pass `true` to server-side function. And this also reminds other engineers who don't know the backend implementation details that they should consider whether just query custom result index is enough.

Got it, yeah I'm ok with this since it does add more flexibility, and easier to change in the future. Just hesitant to add more parameters since some components already have such a long list, and overall prefer to abstract out the details to be determined on server-side and/or downstream functions if possible.

Can you confirm the functionality is working fine for non-custom and custom detectors? If so I'm ok to approve.

ylwu-amzn:
> > > So looks like this adds a param to the following redux functions: `getDetectorResults()` `searchResults()` `getDetectorLiveResults()`
> > > which ends up making calls to the following server-side functions: `searchResults()` `getDetectors()` (which sets the new param as `false` to guarantee results for all detectors) `getAnomalyResults()`
> > > Any reason why the logic for determining whether `onlyQueryCustomResultIndex` should be `true/false` can't be determined within the server-side functions alone? From my understanding, on client-side, passing `resultIndex = ALL_CUSTOM_AD_RESULT_INDICES` always corresponds to `onlyQueryCustomResultIndex = false`, and vice versa. So couldn't the server-side functions determine `onlyQueryCustomResultIndex` by checking the value of `resultIndex.startsWith(CUSTOM_AD_RESULT_INDEX_PREFIX)`? Seems this could minimize the overall code changes, and not add any extra params which have to be passed from client-side -> server-side
> > 
> > 
> > Yeah, we can do that. Here I expose the parameter to make it more flexible for client side. If we want to query only `ALL_CUSTOM_AD_RESULT_INDICES` in future like showing how many anomalies in custom result indices, we can just pass `true` to server-side function. And this also reminds other engineers who don't know the backend implementation details that they should consider whether just query custom result index is enough.
> 
> Got it, yeah I'm ok with this since it does add more flexibility, and easier to change in the future. Just hesitant to add more parameters since some components already have such a long list, and overall prefer to abstract out the details to be determined on server-side and/or downstream functions if possible.
> 
> Can you confirm the functionality is working fine for non-custom and custom detectors? If so I'm ok to approve.

Yeah, I have deployed this PR to our test cluster. All detectors look good.

ohltyler (approved):
LGTM, thanks for the change!

amitgalitz (approved):
LGTM, spent some time understanding change and Tyler's questions/comments above helped a lot. Thanks for adding this!

PR #136, opened 2021-11-10
Fix bug of missing annotations on feature charts

ohltyler:

### Description

When the feature charts are rendered as a child component in the `AnomalyHistory` container, the annotations aren't formatted properly. This is due to one missed place where the annotations weren't converted into 2D arrays, where they should have been as part of #107. The proper conversion is already done when the feature chart is rendered as a child of `AnomaliesChart` container (see [here](https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/public/pages/AnomalyCharts/containers/AnomaliesChart.tsx#L349)). Additionally, the `FeatureChart` wasn't handling the new 2D representation of annotations, so that is fixed as well. Added some clarification comments above that change.

The side effect of this bug is that the annotations were always being passed as an empty array, so no annotations were showing up for the charts.

Confirmed the code changes now show proper annotations for all scenarios:
- Sample anomalies / preview (non-HC, single-HC, multi-HC)
- Real-time (non-HC, single-HC, multi-HC)
- Historical (non-HC, single-HC, multi-HC)
- Zooming
- Multiple features
- Multiple time series

### Check List

### DIFF
--- a/public/pages/AnomalyCharts/components/FeatureChart/FeatureChart.tsx
+++ b/public/pages/AnomalyCharts/components/FeatureChart/FeatureChart.tsx
@@ -37,7 +37,6 @@ import { darkModeEnabled } from '../../../../utils/opensearchDashboardsUtils';
 import {
   prepareDataForChart,
   getFeatureMissingDataAnnotations,
-  filterWithDateRange,
   flattenData,
   convertToEntityString,
 } from '../../../utils/anomalyResultUtils';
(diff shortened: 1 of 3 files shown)


public/pages/AnomalyCharts/components/FeatureChart/FeatureChart.tsx
```diff
@@ -143,17 +142,6 @@ export const FeatureChart = (props: FeatureChartProps) => {
...
-    if (isEmpty(props.annotations)) {
-      return [];
-    }
-    return filterWithDateRange(
-      props.annotations,
-      props.dateRange,
-      'coordinates.x0'
-    );
-  };
-
```
ohltyler, line 156: not needed anymore, since all filtering is done prior to being passed to this component, which should be stateless.

ylwu-amzn (approved):

amitgalitz (approved):

kaituo (approved):

PR #158, opened 2021-12-01
Improve error handling on missing result index errors on detector detail pages

### LINKED ISSUE #145: [BUG] Historical analysis result page shows loading if no readable custom result indices found
If user has no read permission on custom result index, historical result page will show loading state, but the search result response shows `{"ok":false,"error":"No readable custom result indices found"}`

ohltyler:

### Description
Before, if a custom result index was deleted for a detector that was configured with it, then the detector results pages (both real-time and historical) would get stuck in an infinite loading state due to uncaught exceptions on server-side (see #145 for details). 

This improves the experience in two ways:
1. Catching server-side exceptions on client-side to prevent the page from being stuck in infinite loading state
2. Adding an informative callout at the top level of the detector details page indicating a missing result index and next steps to take.

All wording & functionality has been signed off by UX.

Testing
- Confirmed no warnings on non-custom-result-index-detectors
- Confirmed the callout goes away when switching tabs if the index has been recreated in the background
- NOTE: no UT added due to complications in mocking. Will be added in integ tests as a separate effort to improve overall integ test scenarios / coverage.

Screenshots

New callout:

screencapture-localhost-5601-app-anomaly-detection-dashboards-2021-12-14-10_21_53

### Issues Resolved

Closes #145

### Check List

### DIFF
--- a/public/pages/DetectorDetail/containers/DetectorDetail.tsx
+++ b/public/pages/DetectorDetail/containers/DetectorDetail.tsx
@@ -155,7 +155,7 @@ export const DetectorDetail = (props: DetectorDetailProps) => {
   }, []);
 
   // Getting all visible indices. Will re-fetch if changes to the detector (e.g.,
-  // detector starts, result index re-created or user switches tabs to re-fetch detector)
+  // detector starts, result index recreated or user switches tabs to re-fetch detector)
   useEffect(() => {
     const getInitialIndices = async () => {
       await dispatch(getIndices('')).catch((error: any) => {
@@ -405,7 +405,7 @@ export const DetectorDetail = (props: DetectorDetailProps) => {
                 'resultIndex',
                 ''
               )}' has been deleted and all anomaly results have been lost. To start saving anomaly results, 
-              restart real-time or historical detection, or re-create the index.`}
+              restart real-time or historical detection, or recreate the index.`}
               color="danger"
               iconType="alert"
             >
@@ -422,21 +422,21 @@ export const DetectorDetail = (props: DetectorDetailProps) => {
                   await dispatch(createIndex(indexConfig, true))
                     .then(() => {
                       core.notifications.toasts.addSuccess(
-                        `Successfully re-created result index ${get(
+                        `Successfully recreated result index '${get(
                           detector,
                           'resultIndex',
                           ''
-                        )}`
+                        )}'.`
                       );
                     })
                     .catch((error: any) => {
                       console.error(error);
                       core.notifications.toasts.addDanger(
-                        `Error re-creating result index ${get(
+                        `Error recreating result index '${get(
                           detector,
                           'resultIndex',
                           ''
-                        )}`
+                        )}'.`
                       );
                     })
                     .finally(() => {
@@ -444,7 +444,7 @@ export const DetectorDetail = (props: DetectorDetailProps) => {
                     });
                 }}
               >
-                {isCreatingIndex ? 'Re-creating...' : 'Re-create index'}
+                {isCreatingIndex ? 'Recreating...' : 'Recreate index'}
               </EuiButton>
             </EuiCallOut>
           ) : null}
--- a/public/pages/DetectorDetail/containers/DetectorDetail.tsx
+++ b/public/pages/DetectorDetail/containers/DetectorDetail.tsx
@@ -22,14 +22,16 @@ import {
   EuiText,
   EuiFieldText,
   EuiLoadingSpinner,
+  EuiButton,
 } from '@elastic/eui';
 import { CoreStart } from '../../../../../../src/core/public';
 import { CoreServicesContext } from '../../../components/CoreServices/CoreServices';
 import { get, isEmpty } from 'lodash';
 import { RouteComponentProps, Switch, Route, Redirect } from 'react-router-dom';
-import { useDispatch } from 'react-redux';
+import { useDispatch, useSelector } from 'react-redux';
 import { useFetchDetectorInfo } from '../../CreateDetectorSteps/hooks/useFetchDetectorInfo';
 import { useHideSideNavBar } from '../../main/hooks/useHideSideNavBar';
+import { AppState } from '../../../redux/reducers';
 import {
   deleteDetector,
   startDetector,
@@ -37,6 +39,7 @@ import {
   getDetector,
   stopHistoricalDetector,
 } from '../../../redux/reducers/ad';
+import { getIndices, createIndex } from '../../../redux/reducers/opensearch';
 import { getErrorMessage, Listener } from '../../../utils/utils';
 import { darkModeEnabled } from '../../../utils/opensearchDashboardsUtils';
 import { BREADCRUMBS } from '../../../utils/constants';
@@ -53,6 +56,8 @@ import {
   prettifyErrorMessage,
 } from '../../../../server/utils/helpers';
 import { DETECTOR_STATE } from '../../../../server/utils/constants';
+import { CatIndex } from '../../../../server/models/types';
+import { containsIndex } from '../utils/helpers';
 
 export interface DetectorRouterProps {
   detectorId?: string;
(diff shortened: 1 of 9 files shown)


ohltyler:
CI failing due to `1.x` in core Dashboards being bumped to 1.3. Ideally after cutting 1.2 branch, and releasing 1.2, `main` should be bumped to `1.3`.

ohltyler:
> CI failing due to `1.x` in core Dashboards being bumped to 1.3. Ideally after cutting 1.2 branch, and releasing 1.2, `main` should be bumped to `1.3`. Reached out to owners of dependent plugins (common-utils, job-scheduler) to help resolve.

Update: these 2 plugins have been bumped to 1.3. AD and AD Dashboards should be able to bump versions now too

public/pages/DetectorDetail/utils/helpers.tsx
```diff
@@ -107,3 +108,13 @@ export const getDetectorStateDetails = (
     </Fragment>
   );
 };
+
+export const containsIndex = (index: string, indices: CatIndex[]) => {
```
ylwu-amzn, line 112: Is it possible that `indices` is null?

ohltyler: Good callout - I checked redux stores and confirmed the init values are empty arr (`[]`), and can only be overwritten by a valid set of indices from server-side. But just in case something is missed I think it's good to add a null check here - I'll add one

public/pages/DetectorDetail/containers/DetectorDetail.tsx
```diff
@@ -138,6 +154,18 @@ export const DetectorDetail = (props: DetectorDetailProps) => {
     scroll(0, 0);
   }, []);
 
+  // Getting all visible indices. Will re-fetch if changes to the detector (e.g.,
+  // detector starts, result index recreated or user switches tabs to re-fetch detector)
+  useEffect(() => {
+    const getInitialIndices = async () => {
+      await dispatch(getIndices('')).catch((error: any) => {
+        console.error(error);
+        core.notifications.toasts.addDanger('Error getting all indices');
```
ylwu-amzn, line 163: If user has no permission to list indices, what error message will show up?

ohltyler: It would be whatever is returned from the backend, like the regular permissions errors. The reason we only have a generic message in the toast is tor prevent some verbose / not formatted message show up which could add confusion. But it's logged on client-side console as well as server-side (and since it's permissions-related, would be logged in backend also)

public/pages/DetectorDetail/containers/DetectorDetail.tsx
```diff
@@ -365,6 +393,61 @@ export const DetectorDetail = (props: DetectorDetailProps) => {
...
+                style={{ marginLeft: '24px' }}
+                fill={false}
+                isLoading={isCreatingIndex}
+                onClick={async () => {
+                  setIsCreatingIndex(true);
+                  const indexConfig = {
+                    index: get(detector, 'resultIndex', ''),
+                    body: {},
+                  };
+                  await dispatch(createIndex(indexConfig, true))
```
ylwu-amzn, line 422: Will we use default AD result index mapping to re-create custom result index?

ohltyler: Currently it's pulling the mappings from [here](https://raw.githubusercontent.com/opensearch-project/anomaly-detection/main/src/main/resources/mappings/anomaly-results.json). More details on that [here](https://github.com/ohltyler/anomaly-detection-dashboards-plugin-1/blob/result-index-handling/server/routes/opensearch.ts#L185).

public/pages/DetectorDetail/containers/DetectorDetail.tsx
ylwu-amzn: Cool, it's ok to show regular permissions errors.

public/redux/reducers/opensearch.ts
```diff
@@ -295,36 +297,35 @@ export const deleteIndex = (index: string): APIAction => ({
...
+    if (exactMatchedAliases.length || exactMatchedIndices.length) {
+      //If we have exact match just return that
       return {
-        indices: partialMatchedIndices,
-        aliases: partialMatchedAliases,
+        indices: exactMatchedIndices,
+        aliases: exactMatchedAliases,
       };
+    } else {
+      //No results found for exact match, append wildCard and get partial matches if exists
```
ylwu-amzn, line 316: Why we need to get partial matched indices?

server/routes/opensearch.ts
```diff
@@ -177,10 +178,45 @@ export default class OpenSearchService {
...
+    let { isResultIndex } = request.params as { isResultIndex: any };
+    isResultIndex = JSON.parse(isResultIndex) as boolean;
+    let anomalyResultIndexMappings = {};
+
+    // If creating a custom result index: need to fetch the latest anomaly result index mappings.
+    // Currently pulling the stored JSON mappings from backend source code. Prefer to
+    // pull from backend rather than persisting a copy of the latest mappings in frontend code.
+    if (isResultIndex) {
+      try {
+        await fetch(ANOMALY_RESULTS_INDEX_MAPPINGS_LINK)
```
ylwu-amzn, line 190: Not sure if this will bring any security/performance issue, but seems AD plugin will be impacted if Github crashes and AD will run slow if Github has performance issue. Maybe we can either copy the mapping to frontend package or create a new API in backend if don't want to maintain duplicate index mappings in both frontend&backend package?

public/redux/reducers/opensearch.ts
ohltyler: There isn't any changed code here - seems it was initially saved without being formatted. The only changed lines in this file are lines 277-283.

server/routes/opensearch.ts
ohltyler: Yeah still playing around with the correct approach here. In fact, I just realized this would break in future versions, since the link is tied to `main`. If we change the mappings in 1.4 for example, but are running Dashboards 1.3, then it would be expecting the mappings associated with 1.3, but would get the ones from 1.4, since it's pulling from `main`.

We would have a similar problem with the API approach, unless we added a param to specify the schema version - but that still doesn't solve the problem of needing to make frontend changes every time the result index schema is changed (e.g., would need to change the schema version in api call). 

Also, I think it's preferred to not add APIs if only for this very small use-case, so to me it seems the best approach could be persisting a copy in the frontend. That way the correct mappings can be coupled with the running version of the cluster. But even this may not be maintainable, and could be forgotten to be updated if more backend schema changes happen.

What do you think? I can sync with UX again to possibly remove this functionality so we don't risk it breaking in the future.

ylwu-amzn: I think another option is to just show some warning message like "Your detector is using custom result index: xxx, but it's not found in cluster. The index will be re-created when you start realtime job or historical analysis.".
So we don't provide re-create result index button. If user don't need to restart realtime/historical analysis, they actually don't need the custom result index. If they need to restart, AD backend will help recreate before starting realtime/historical job.

public/pages/DetectorDetail/containers/DetectorDetail.tsx
amitgalitz: What does it mean by fetching the latest anomaly result index mapping? Is this different than the initial default mapping? Also what would the name be of the recreated custom result index.

server/routes/opensearch.ts
ohltyler: Agreed, wording along the lines of that would be more clear and allow us to remove the create index button. Regardless I think it's agreed we don't want to persist mappings logic in the frontend at all.

public/pages/DetectorDetail/containers/DetectorDetail.tsx
ohltyler: It would be the default mapping based on the current version. But since the mappings have changed over different versions (like adding high-cardinality-related fields to the results), the idea is to get the corresponding version of the mappings. The name would be the stored custom result index for the detector.

Either way this functionality will probably be removed and we will let the backend init the index with the correct mappings.

ohltyler:
Converting to draft until a better approach is decided on.

server/routes/opensearch.ts
ohltyler: @ylwu-amzn UX has agreed to go with this option. I'll remove the create index button and just tune the wording in the callout.

ylwu-amzn: Thanks @ohltyler , will review once new commit published.

public/pages/DetectorDetail/containers/DetectorDetail.tsx
ohltyler: This is removed now.

ohltyler:
@opensearch-project/opensearch-ux please have a look at the wording in the screenshot in the PR description to confirm. Thank you! @kgcreative is familiar with the current efforts

server/routes/opensearch.ts
ohltyler: New changes have been pushed.

kgcreative:
Thanks @ohltyler! Visual changes look good to me!

ylwu-amzn (approved):
LGTM. Thanks for the change!

amitgalitz (approved):
LGTM, thanks for changing this!

Issue #117, opened 2021-11-05
Add workflow to run E2E tests with security enabled 


amitgalitz:
Currently the workflow for E2E testing is not integrated with the security plugin meaning e2e tests are only run with security disabled. We need to add a workflow that runs e2e tests with security enabled.

PR #163, opened 2021-12-17
Add remote integ test workflow; clean up old integ test workflow

### LINKED ISSUE #162: Run AD tests from opensearch-dashboards-functional-test repo
See https://github.com/opensearch-project/opensearch-dashboards-functional-test/issues/65 for details.

ohltyler:

### Description

Adds a CI workflow where the integ tests that are stored in opensearch-dashboards-functional-test repo are pulled and ran. Note only a few tests have been added there so far (see [here](https://github.com/opensearch-project/opensearch-dashboards-functional-test/pull/64)) - the long-term plan will be to migrate all integ tests there. Also cleans up the existing CI workflow which incorrectly states that it's running with security enabled.

Note that the added workflow is for testing against the upcoming release version, thus the cluster setup process is complex and limited. Because there is no snapshot bundled builds, we rely on the existing methods:
1. Set up OpenSearch by running `./gradlew run` in backend AD repo, which spins up a cluster with only AD plugin installed (core OpenSearch is pulled via `min` artifact).
2. Set up Dashboards by pulling and running from source, with only AD plugin installed.
3. No security plugin (and no security-enabled tests) due to complications in installing it alongside AD plugin.

### Issues Resolved

Closes #162 

### Check List

### DIFF
--- a/.github/workflows/remote-integ-tests-workflow.yml
+++ b/.github/workflows/remote-integ-tests-workflow.yml
@@ -91,4 +91,4 @@ jobs:
         uses: cypress-io/github-action@v2
         with:
           working-directory: opensearch-dashboards-functional-test
-          command: yarn run cypress run --env SECURITY_ENABLED=false --spec "cypress/integration/plugins/anomaly-detection-dashboards-plugin/*.js"
+          command: yarn run cypress run --env SECURITY_ENABLED=false --spec cypress/integration/plugins/anomaly-detection-dashboards-plugin/*.js
--- a/.github/workflows/remote-integ-tests-workflow.yml
+++ b/.github/workflows/remote-integ-tests-workflow.yml
@@ -1,6 +1,7 @@
-# Soon-to-be deprecated way of running integ tests. In the future all tests will be moved to
-# https://github.com/opensearch-project/opensearch-dashboards-functional-test
-name: E2E tests workflow
+# Running AD integ tests stored in https://github.com/opensearch-project/opensearch-dashboards-functional-test
+# In the future we should pull dependencies from bundled build snapshots. Because that is not available
+# yet we build the cluster from source (besides core Opensearch, which is a pulled min artifact).
+name: Remote integ tests workflow
 on:
   push:
     branches:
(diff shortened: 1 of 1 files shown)


tianleh (approved):

.github/workflows/e2e-tests-workflow.yml
```diff
@@ -1,3 +1,5 @@
+# Soon-to-be deprecated way of running integ tests. In the future all tests will be moved to
+# https://github.com/opensearch-project/opensearch-dashboards-functional-test
```
ylwu-amzn, line 2: After moving to `opensearch-dashboards-functional-test`, we should clone `opensearch-dashboards-functional-test` to local to run E2E test? Suggest update readme once code ready in `opensearch-dashboards-functional-test`

.github/workflows/e2e-tests-workflow.yml
```diff
@@ -10,7 +12,7 @@ env:
   OPENSEARCH_DASHBOARDS_VERSION: '1.x'
   OPENSEARCH_VERSION: '1.3.0-SNAPSHOT'
 jobs:
-  test-with-security:
+  test-without-security:
```
ylwu-amzn, line 15: Don't see `test-with-security`, so we don't support e2e with security enabled? How about add some TODO ?

.github/workflows/e2e-tests-workflow.yml
ohltyler: @ylwu-amzn yes - and good callout about the README. We will need to update once all tests have moved over there, and once we have removed them from this codebase.

.github/workflows/e2e-tests-workflow.yml
ohltyler: Right, it's a known issue tracked in #117. Ideally the security snapshot artifacts get published to maven (similar to common-utils) and they can be pulled in and built in a new workflow.

ohltyler: The previous solution was using snapshot docker images but those are only built for previous versions rather than upcoming versions. There is future infra plans to have these available for the upcoming versions which will eventually resolve all of these build-related issues.

ylwu-amzn (approved):
LGTM. Thanks!

kaituo (approved):

PR #168, opened 2022-01-03
Reformat using Prettier CLI and add doc to developer guide

### LINKED ISSUE #111: Add documentation on formatting in README
Auto-formatting can be done in a few different ways, these should be added to the README somewhere:
1. Using VS code - there is a "Prettier - Code formatter" extension, and coupled with "Format on save" set to true, allows for auto-formatting whenever saving files.
2. Using CLI - there is a `prettier` yarn dependency that can format everything via running `yarn prettier <file-to-format> --write` in base plugin repo.

amitgalitz:

### Description

* Added documentation to developer guide explaining that the code base uses prettier for any code formatting. Developers should use the Prettier CLI to format code instead of IDE plugins. Some extensions such as VS code added a new setting for the plugin to use the Prettier version defined in the project modules, however this still didn't work on my end. If developers still try to use an IDE Prettier plugin then they should be responsible in checking what Prettier version is used.  
* I also run the `yarn prettier --write .` command to reformat any files that haven't been reformatted in the past with Prettier. I made sure to check each file was only change through formatting.
* Added `.yml` and `.md` files to prettier ignore as overall style and formatting of those files are based on opensearch-dashboard. Specifically this caused issues with release notes files. 

### Issues Resolved

#111 

### Check List


DEVELOPER_GUIDE.md
```diff
@@ -69,3 +69,14 @@ In the base OpenSearch Dashboards directory, run
 
   Runs end-to-end tests on a currently running OpenSearch Dashboards server. Defaults to run the tests on `localhost:5601`, although you can change this to run on any
   OpenSearch Dashboards server with the command `yarn cy:run --config baseUrl=<your-custom-URL>`
+
+
+### Formatting
```
ohltyler, line 74: Can you sync with core dashboards folks about adding something there, and then plugins can copy/reference it in their respective repos? Just thinking of how to make it more standard across the dashboards plugins.

.prettierignore
```diff
@@ -0,0 +1,2 @@
+*.md
```
ohltyler, line 1: Prettier does help a lot in auto-formatting markdown files and re-organizing / creating links at the top of READMEs, etc. Are you sure we want to exclude it? If the only issue is related to the release notes, can you change the glob pattern to only ignore release notes md files?

DEVELOPER_GUIDE.md
amitgalitz: Opensearch-Dashboards does have a section on Prettier (https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/DEVELOPER_GUIDE.md#prettier-and-linting). It is mainly mentioning the move to Prettier for code formatting and using Prettier and ESLint rules over any other rules. I'll reach out regarding CLI vs plugin and if they ever had any discrepancy issue.

.prettierignore
amitgalitz: Only major issue seems to be release-notes. Other thing was that I knew we had some documentation that was based of Opensearch-Dashboards and since there were issues with release-notes I assumed other issues would exist. Doesn't seem like it after further checking so I'll change it to only ignore release notes md files.

ohltyler: Gotcha, yeah that makes sense. I think we're ok to just ignore release notes for now, since that is the only md file that will cause workflows to break if formatting is off.

DEVELOPER_GUIDE.md
ohltyler: Oh cool, sounds good. If they have no preference I think it's fine to leave what you've added.

amitgalitz: synced up with core dashboards and they mentioned they also noticed discrepancies and they just rely on CLI for final checks pre-PR but don't really have a preference as long as code formatting uses prettier.

ohltyler (dismissed):
LGTM, thanks for adding! Seems the CI failures are still related to the backend plugin so no concerns from me there.

ohltyler:
@amitgalitz Retried CI and looks good now

ylwu-amzn (approved):
LGTM. Thanks for the change!

amitgalitz:
> @amitgalitz can you explain what the force-push was for? Was there any changes from my last review?

Just rebased with main, realized all that is the 1 line of code change from fixing heatmap PR but just wanted to stay consistent

ohltyler (approved):

jackiehanyang (approved):

PR #178, opened 2022-02-14
Fix custom expression filter cannot show

kaituo:
### Description

This PR fixes that a new detector with custom filter created using API cannot show.

Testing done:
1. Created a detector with a custom filter using API and verified the change fixed the issue.

Before:
Screen Shot 2022-02-12 at 10 38 55 AM

After:
Screen Shot 2022-02-12 at 11 07 12 AM

Screen Shot 2022-02-12 at 11 07 59 AM

### Check List

  - [X] All tests pass

### DIFF
--- a/public/pages/ReviewAndCreate/components/FilterDisplayList/FilterDisplayList.tsx
+++ b/public/pages/ReviewAndCreate/components/FilterDisplayList/FilterDisplayList.tsx
@@ -26,11 +26,14 @@ export const FilterDisplayList = (props: FilterDisplayListProps) => {
   const [filterIndex, setFilterIndex] = useState<number>(-1);
   let filters = get(props, 'uiMetadata.filters', []);
   const oldFilterType = get(props, 'uiMetadata.filterType', undefined);
-  const isOldDetector = !isEmpty(oldFilterType);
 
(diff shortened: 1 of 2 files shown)


public/pages/ReviewAndCreate/components/FilterDisplayList/FilterDisplayList.tsx
```diff
@@ -30,7 +30,7 @@ export const FilterDisplayList = (props: FilterDisplayListProps) => {
 
   // Old detectors with custom filters will have no filter list, but
   // will have a populated filter query
-  if (isEmpty(filters) && isOldDetector && !isEmpty(props.filterQuery)) {
```
ohltyler, line 33: Looks like you can actually remove the `isOldDetector` check entirely. If it's true, it's from an old detector that's been created/updated on old UI. If it's false, it's only been created/updated from API (no `uiMetadata.*` fields). Regardless of either of those situations, we want to show the custom filter if `props.filterQuery` isn't empty.

amitgalitz (approved):
LGTM, thanks for fixing this!

ohltyler (approved):
Thanks for catching and fixing this bug!

PR #202, opened 2022-03-07
Non blocker

amitgalitz:

### Description

This PR integrates the new additional type parameter for validation API that includes `model` validation. Since `model` (non-blocker) validation also runs `detector` (blocker) validation, currently I only call the validate API with `model` parameter. In the future we can separate the two if we want to make the non-blocker validation an optional action. With this API integrated the user will be able to see any non-blocking issues.

#### Changes include:
* If response type is model then object passed to model or detector definition will include that information and the callout will be of color `warning` instead of `danger`

### Screenshot:

Screen Shot 2022-03-07 at 8 22 12 AM

### Check List

### DIFF
--- a/public/pages/ReviewAndCreate/components/DetectorDefinitionFields/DetectorDefinitionFields.tsx
+++ b/public/pages/ReviewAndCreate/components/DetectorDefinitionFields/DetectorDefinitionFields.tsx
@@ -28,7 +28,7 @@ import {
 import { FilterDisplayList } from '../FilterDisplayList';
 import { ConfigCell, FixedWidthRow } from '../../../../components/ConfigCell';
 import { toStringConfigCell } from '../../utils/helpers';
-
+import { isEqual } from 'lodash';
 interface DetectorDefinitionFieldsProps {
   detector: Detector;
   onEditDetectorDefinition(): void;
@@ -94,19 +94,36 @@ export const DetectorDefinitionFields = (
         !props.validDetectorSettings &&
         props.validationResponse.hasOwnProperty('message')
       ) {
-        return (
-          <EuiCallOut
-            title="Issues found in the detector settings"
-            color="danger"
-            iconType="alert"
-            size="s"
-            style={{ marginBottom: '10px' }}
-          >
-            <ul>
-              <li>{props.validationResponse.message}</li>
-            </ul>
-          </EuiCallOut>
-        );
+        if (isEqual(props.validationResponse.validationType, 'model')) {
+          return (
+            <EuiCallOut
+              title="We identified some areas that might improve your model"
+              color="warning"
+              iconType="iInCircle"
+              size="s"
+              style={{ marginBottom: '10px' }}
+            >
+              {JSON.stringify(props.validationResponse.message).replace(
+                /\"/g,
+                ''
+              )}
+            </EuiCallOut>
+          );
+        } else {
+          return (
+            <EuiCallOut
+              title="Issues found in the detector settings"
+              color="danger"
+              iconType="alert"
+              size="s"
+              style={{ marginBottom: '10px' }}
+            >
+              <ul>
+                <li>{props.validationResponse.message}</li>
+              </ul>
+            </EuiCallOut>
+          );
+        }
       } else {
         return null;
       }
--- a/public/pages/ReviewAndCreate/components/DetectorDefinitionFields/DetectorDefinitionFields.tsx
+++ b/public/pages/ReviewAndCreate/components/DetectorDefinitionFields/DetectorDefinitionFields.tsx
@@ -20,7 +20,7 @@ import {
   EuiText,
 } from '@elastic/eui';
 import React from 'react';
-import { get } from 'lodash';
+import { get, isEqual } from 'lodash';
 import {
   Detector,
   ValidationSettingResponse,
@@ -28,7 +28,6 @@ import {
 import { FilterDisplayList } from '../FilterDisplayList';
 import { ConfigCell, FixedWidthRow } from '../../../../components/ConfigCell';
 import { toStringConfigCell } from '../../utils/helpers';
-import { isEqual } from 'lodash';
 interface DetectorDefinitionFieldsProps {
   detector: Detector;
   onEditDetectorDefinition(): void;
@@ -94,7 +93,9 @@ export const DetectorDefinitionFields = (
         !props.validDetectorSettings &&
         props.validationResponse.hasOwnProperty('message')
       ) {
-        if (isEqual(props.validationResponse.validationType, 'model')) {
+        if (
+          isEqual(get(props, 'validationResponse.validationType', ''), 'model')
+        ) {
           return (
             <EuiCallOut
               title="We identified some areas that might improve your model"
--- a/public/pages/ReviewAndCreate/components/ModelConfigurationFields/ModelConfigurationFields.tsx
+++ b/public/pages/ReviewAndCreate/components/ModelConfigurationFields/ModelConfigurationFields.tsx
@@ -26,7 +26,7 @@ import {
   FeatureAttributes,
   ValidationModelResponse,
 } from '../../../../models/interfaces';
-import { get, sortBy } from 'lodash';
+import { get, sortBy, isEqual } from 'lodash';
 import ContentPanel from '../../../../components/ContentPanel/ContentPanel';
 import { CodeModal } from '../../../../components/CodeModal/CodeModal';
 import { AdditionalSettings } from '../AdditionalSettings/AdditionalSettings';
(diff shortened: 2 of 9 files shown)


public/pages/ReviewAndCreate/components/DetectorDefinitionFields/DetectorDefinitionFields.tsx
```diff
@@ -28,7 +28,7 @@ import {
 import { FilterDisplayList } from '../FilterDisplayList';
 import { ConfigCell, FixedWidthRow } from '../../../../components/ConfigCell';
 import { toStringConfigCell } from '../../utils/helpers';
-
+import { isEqual } from 'lodash';
```
ohltyler, line 31: nit: combine the import with `get` on line 23

ohltyler: Unfortunately `prettier` doesn't handle this automatically

public/pages/ReviewAndCreate/components/DetectorDefinitionFields/DetectorDefinitionFields.tsx
```diff
@@ -94,19 +94,36 @@ export const DetectorDefinitionFields = (
...
-            iconType="alert"
-            size="s"
-            style={{ marginBottom: '10px' }}
-          >
-            <ul>
-              <li>{props.validationResponse.message}</li>
-            </ul>
-          </EuiCallOut>
-        );
+        if (isEqual(props.validationResponse.validationType, 'model')) {
```
ohltyler, line 97: `props.validationResponse` is listed as an optional field. To prevent an NPE, suggest to change `props.validationResponse.validationType` -> `get(props, 'validationResponse.validationType', '')`

public/pages/ReviewAndCreate/components/DetectorDefinitionFields/DetectorDefinitionFields.tsx
```diff
@@ -94,19 +94,36 @@ export const DetectorDefinitionFields = (
...
+          return (
+            <EuiCallOut
+              title="We identified some areas that might improve your model"
+              color="warning"
+              iconType="iInCircle"
+              size="s"
+              style={{ marginBottom: '10px' }}
+            >
+              {JSON.stringify(props.validationResponse.message).replace(
```
ohltyler, line 106: For consistency, should this be in a bullet-point format like it is for blocker issues on lines 121-123?

ohltyler: Or will only one possible message be returned per non-blocker failure?

public/pages/ReviewAndCreate/components/ModelConfigurationFields/ModelConfigurationFields.tsx
```diff
@@ -253,26 +253,45 @@ export const ModelConfigurationFields = (
...
+            {JSON.stringify(props.validationFeatureResponse.message).replace(
+              /\"/g,
+              ''
+            )}
+          </EuiCallOut>
+        );
+      } else {
+        return (
+          <EuiCallOut
+            title="issues found in the model configuration"
```
ohltyler, line 274: `issues` -> `Issues`

public/pages/ReviewAndCreate/components/ModelConfigurationFields/ModelConfigurationFields.tsx
```diff
@@ -253,26 +253,45 @@ export const ModelConfigurationFields = (
...
-          {props.validationFeatureResponse.hasOwnProperty('sub_issues') ? (
-            getFeatureValidationCallout(props.validationFeatureResponse)
-          ) : (
-            <ul>
-              <li>{JSON.stringify(props.validationFeatureResponse.message)}</li>
-            </ul>
-          )}
-        </EuiCallOut>
-      );
+      if (isEqual(props.validationFeatureResponse.validationType, 'model')) {
```
ohltyler, line 256: Recommend moving lines 256-294 to a helper file in `ReviewAndCreate` so it can be used by `DetectorDefinitionFields` component as well. Looks like the callout logic is duplicated there, with the only difference being the title of the blocker callout - either `"Issues found in the detector settings"` or `"issues found in the model configuration"`

public/pages/ReviewAndCreate/containers/ReviewAndCreate.tsx
```diff
@@ -92,28 +91,45 @@ export function ReviewAndCreate(props: ReviewAndCreateProps) {
...
-                resp.response.detector[issueType].message;
+                resp.response[validationType][issueType].message;
               const detectorSettingIssue: ValidationSettingResponse = {
                 issueType: issueType,
                 message: validationMessage,
+                validationType: validationType,
               };
+
+              // Potentially remove warning toast in these cases
+              if (issueType == 'aggregation' || issueType == 'timeout') {
```
ohltyler, line 118: Why would we want to remove the warning toast here? What does an `aggregation` issue type mean?

public/pages/ReviewAndCreate/components/DetectorDefinitionFields/DetectorDefinitionFields.tsx
amitgalitz: Only one possible message will be returned per non-blocker failure as of now so thought it looks nicer without the bullet point

public/pages/ReviewAndCreate/components/ModelConfigurationFields/ModelConfigurationFields.tsx
amitgalitz: ModelConfigurationFields has additional logic in how it displays the message for any blocker issues. This is due to needing to parse the sub issues for feature attributes issues that are also of type detector/blocker. I could refactor it but it would also mean adding more if checks inside depending if the issue response is feature response from `ModelConfigurationsFields` or detector settings from `DetectorDefinitionFields`. I'll try to see if its less messy to refactor and added the checks or to leave it as is

public/pages/ReviewAndCreate/containers/ReviewAndCreate.tsx
amitgalitz: The reason for this is due to the fact that if there is an `aggregation` issue then it might be an issue that arouse from a change in core and user won't have much they can do about it and for timeout they also can't really re-try since action happens automatically when entering `ReviewAndCreate` page so I thought that I could just make it basically invisible to users instead of additionally giving a warning toast that will confuse them on there next action.

public/pages/ReviewAndCreate/components/ModelConfigurationFields/ModelConfigurationFields.tsx
ohltyler: Ok, missed that separate logic. Probably fine to leave separate if there's too much unique logic per section.

public/pages/ReviewAndCreate/containers/ReviewAndCreate.tsx
ohltyler: In this same `if` block, I see it sets the variables indicating valid detector settings and model configs, when in fact that might not be true if there was a problem performing the queries, or if it timed out. It seems like a bad user experience to see it showing as valid in the callouts, then failing later when actually creating the detector, or trying to start it.

How about having a separate callout text indicating that validation failed to run? It could be generic so that it covers both aggregation and timeout failures. I think showing the toast with more details on the failure (aggregation or timeout) could be helpful to the user, and would allow for quick user feedback in case there is something wrong with the AD code that we would need to fix (e.g., aggregation failure where the queries being made in validation API call need to be updated after core changes something internal).

ohltyler: As discussed separately, these issue types only come up during non-blocker validation, after blocker validation has passed. So in that case it is mostly validated, just not fully validated / not able to provide any recommendations in case there is data sparsity due to any of the configs / features.

I think the 2 options are (1) hide these failures like you're doing now, or (2) find a way to tell the user that blocking validation passed, but non-blocking validation failed to run properly. For simplicity, I think its ok to go with option 1 for this release, and sync with UX team on a solution to option 2 later on.

Suggest to update the comments on line 117 to reflect something along the lines of this

public/pages/ReviewAndCreate/containers/ReviewAndCreate.tsx
```diff
@@ -93,6 +92,7 @@ export function ReviewAndCreate(props: ReviewAndCreateProps) {
   useEffect(() => {
     dispatch(validateDetector(formikToDetector(props.values), 'model'))
       .then((resp: any) => {
+        console.log('resp: ' + JSON.stringify(resp));
```
ohltyler, line 95: can remove this :)

ohltyler (dismissed):
LGTM! Just one leftover log to clean: https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/202#discussion_r821210339

ohltyler (approved):

kaituo (approved):

ohltyler:
Let's wait for the [backend PR](https://github.com/opensearch-project/anomaly-detection/pull/384) to merge before merging this one.

PR #262, opened 2022-05-17
Wait for detector to load before checking indices exist

### LINKED ISSUE #251: [BUG] a lot of getting all indices error when open sample detector page
**What is the bug?**
See a lot of "Error getting all indices" error when open sample detector.

**How can one reproduce the bug?**
1. Create multi node cluster with CDK
2. Create 3 sample detectors
3. Open `sample-ecommerce-detector` 
4. See a lot of `Error getting all indices` error. Open browser network, see many requests of `_indices?index=`.

**What is the expected behavior?**
Just throw one error and don't sent out so many request in short time.

**What is your host/environment?**
 - OS: Amazon Linux 2
 - Version 1.3.1
 - Plugins

**Do you have any screenshots?**
Screen Shot 2022-04-25 at 5 30 40 PM

**Do you have any additional context?**
Add any other context about the problem.

amitgalitz:

### Description
Added two new conditions to two different `useEffect` calls on `DetectorDetail` page. 
First change is to wait for getDetector request to finish loading before checking if the indices themselves exist. Currently we are checking if the indices exist on every change to the detector and while the detector is still loading it continuously loops this API call to check for indices. This then triggers a change in AppState which triggers the change in `detector` variable hence the loop.

The second change is under the same idea where we don't need to call the `useEffect` that calls on `hideStopDetectorModal` if the detector itself is empty.

Further improvements will be made in later PR to optimize `useFetchDetectorInfo` further which is the call updating the `detector` variable.

### Issues Resolved
resolves #251 

### Check List

### DIFF
--- a/public/pages/DetectorDetail/containers/DetectorDetail.tsx
+++ b/public/pages/DetectorDetail/containers/DetectorDetail.tsx
@@ -161,7 +161,10 @@ export const DetectorDetail = (props: DetectorDetailProps) => {
         core.notifications.toasts.addDanger('Error getting all indices');
       });
     };
-    getInitialIndices();
+    // only need to check if indices exist after detector finishes loading
+    if (!isLoadingDetector) {
+      getInitialIndices();
+    }
   }, [detector]);
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


ohltyler (approved):
LGTM - thanks for these improvements!

ohltyler:
@amitgalitz we want to backport to 2.0 as well right?

amitgalitz:
> @amitgalitz we want to backport to 2.0 as well right?

yes and 1.x, 1.3, forgot to add labels, just added them

ylwu-amzn (approved):

Issue #223, opened 2022-03-31
[BUG]Missing data call out if window delay is large


kaituo:
**Describe the bug**
When I configure window delay to 20 minutes, the missing data call out will show. My suspicion is if window delay is more than certain number, this will always show. Might be a bug of missing data calculation.

**To Reproduce**
Create/run a detector with window delay 20 minutes.

**Expected behavior**
This should not show.

**Screenshots**
![Screen Shot 2022-03-31 at 10 25 24 AM](https://user-images.githubusercontent.com/5303417/161114417-0dcd66d6-ea8d-4213-a991-0ba8e5a8a8d7.png)

![Screen Shot 2022-03-31 at 10 26 46 AM](https://user-images.githubusercontent.com/5303417/161114523-401d4e91-1ae4-4376-83fb-caf17d62a68d.png)

ohltyler:
The missing feature callout only shows when fetching feature data points based on the interval and time range, and if there is any empty points. My assumption is that the window delay offset isn't being considered in the time range used in the feature fetching query, or in the displayed chart, such that it shows as empty, when it's actually expected to be empty based on the window delay. But, I'm unable to reproduce with a variety of window delays (20 min, 25 min, 30 min), with an interval of 10 mins. What interval were you using here?

kaituo:
I am using 5 min interval.

kaituo:
The issue is that we didn't consider window delay when checking missing data. The fix is to include window delay in missing feature calculations.

PR #277, opened 2022-06-17
Fix missing data call out if window delay is large

kaituo:
### Description

The missing feature callout only shows when fetching feature data points based on the interval and time range, and if there are any empty points. Previously, when I configured the window delay to 20 minutes, the missing data call out will show. The reason is that the window delay offset isn't being considered in the time range used in the feature fetching query, or in the displayed chart, such that it shows as empty, when it's actually expected to be empty based on the window delay.This PR fixed the bug by considering window delay in missing feature calculations.

Testing done:
1. done e2e testing and verified the problem is fixed.
2. added unit tests.

Note: since I found the issue in an OS 1.1 domain, I started fixing in the 1.1 branch so that I can verify if the problem is fixed. Will forward port to later versions.

### Issues Resolved

https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/issues/223

### Check List

  - [x] All tests pass

### DIFF
--- a/public/models/interfaces.ts
+++ b/public/models/interfaces.ts
@@ -27,6 +27,8 @@
 import { InitProgress } from '../../server/models/interfaces';
 import { DATA_TYPES } from '../utils/constants';
 import { DETECTOR_STATE } from '../../server/utils/constants';
+import { Duration } from 'moment';
+import moment from 'moment';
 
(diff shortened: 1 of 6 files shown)


kaituo:
close since Mend security Check seems stuck.

PR #278, opened 2022-06-18
Fix missing data call out if window delay is large

kaituo:
### Description

The missing feature callout only shows when fetching feature data points based on the interval and time range, and if there are any empty points Previously, when I configured the window delay to 20 minutes, the missing data call out will show. The reason is that the window delay offset isn't being considered in the time range used in the feature fetching query, or in the displayed chart, such that it shows as empty, when it's actually expected to be empty based on the window delay.This PR fixed the bug by considering window delay in missing feature calculations.

Testing done:
1. done e2e testing and verified the problem is fixed.
2. added unit tests.

### Issues Resolved

https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/issues/223

### Check List

  - [x] All tests pass

### DIFF
--- a/public/pages/DetectorResults/containers/AnomalyResults.tsx
+++ b/public/pages/DetectorResults/containers/AnomalyResults.tsx
@@ -208,11 +208,11 @@ export function AnomalyResults(props: AnomalyResultsProps) {
   const checkLatestFeatureDataPoints = async () => {
     let windowDelayInMinutes = 0;
     if (detector.windowDelay !== undefined) {
-      let windowDelay = detector.windowDelay.period;
-      let windowDelayUnit = get(windowDelay, 'unit',  UNITS.MINUTES);
+      const windowDelay = detector.windowDelay.period;
+      const windowDelayUnit = get(windowDelay, 'unit',  UNITS.MINUTES);
 
       // current time minus window delay
-      let windowDelayInterval = get(windowDelay, 'interval',  0);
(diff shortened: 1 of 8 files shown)


public/pages/DetectorResults/containers/AnomalyResults.tsx
```diff
@@ -204,10 +204,29 @@ export function AnomalyResults(props: AnomalyResultsProps) {
...
 
+  // ran during componentDidMount and componentDidUpdate
   const checkLatestFeatureDataPoints = async () => {
+    let windowDelayInMinutes = 0;
+    if (detector.windowDelay !== undefined) {
+      let windowDelay = detector.windowDelay.period;
+      let windowDelayUnit = get(windowDelay, 'unit',  UNITS.MINUTES);
+
+      // current time minus window delay
+      let windowDelayInterval = get(windowDelay, 'interval',  0);
```
ohltyler, line 215: nit: these 3 can be `const`s

ohltyler (approved):
Overall LGTM! Thanks for adding documentation on the util fns and fixing the bug.

public/pages/DetectorResults/containers/AnomalyResults.tsx
kaituo: changed

amitgalitz (approved):
LGTM! thanks for the bug fix and added tests

ohltyler (approved):

PR #329, opened 2022-10-24
bumping main to 3.0

amitgalitz:

### Description

bumping main to 3.0 branch

### Issues Resolved

https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/issues/290

### Check List

### DIFF
--- a/.github/workflows/remote-integ-tests-workflow.yml
+++ b/.github/workflows/remote-integ-tests-workflow.yml
@@ -10,8 +10,8 @@ on:
     branches:
       - "*"
 env:
-  OPENSEARCH_DASHBOARDS_VERSION: '2.x'
-  OPENSEARCH_VERSION: '2.3.0-SNAPSHOT'
+  OPENSEARCH_DASHBOARDS_VERSION: 'main'
+  OPENSEARCH_VERSION: '3.0.0-SNAPSHOT'
 jobs:
   test-without-security:
     name: Run integ tests without security
--- a/opensearch_dashboards.json
+++ b/opensearch_dashboards.json
@@ -1,7 +1,7 @@
 {
   "id": "anomalyDetectionDashboards",
-  "version": "2.3.0.0",
-  "opensearchDashboardsVersion": "2.3.0",
+  "version": "3.0.0.0",
+  "opensearchDashboardsVersion": "3.0.0",
   "configPath": ["anomaly_detection_dashboards"],
   "requiredPlugins": ["navigation"],
   "optionalPlugins": [],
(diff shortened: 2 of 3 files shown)


ohltyler (approved):

amitgalitz:
Remote integ tests are flaky here: 
Run 1: 3 of 8 test suites had failures
Run 2 :  4 of 8 test suites had failures
Run 3: It couldn't find job-scheduler 3.0.0 zip even though it was found for previous 2 runs

amitgalitz:
Job scheduler is currently failing for 3.x so we can't successfully run CI here: https://github.com/opensearch-project/job-scheduler/issues/249
On a side note: In my opinion we really should be careful how much time we decide to spend to get main stable cause if we are constantly having breaking changes from core and other dependencies it can create a constant loop of maintenance that will slow down our development speed.

ohltyler:
> 

I think it's worth keeping an eye out on the amount of effort needed to remain compatible with the next major version / 3.0. The main idea is that it's supposed to be the 'cutting edge' branch that will have all of the newest changes in it, and it eases the amount of effort we will have to make during a 3.0 release since we will have already done the work to be compatible with it based on all of the breaking changes in it as they are merged. But 3.0 shouldn't consistently break - if so, it's worth further discussion and probably opening a ticket in  https://github.com/opensearch-project/opensearch-plugins

ohltyler:
@amitgalitz actually looks like there's a good thread about this here: https://github.com/opensearch-project/opensearch-plugins/issues/142#issuecomment-1289727536

amitgalitz:
The remote tests have been extra flaky on this PR. The create AD test suite is failing more consistently with this error: `Expected to find element: [data-test-subj="defineOrEditDetectorTitle"], but never found it.`
Looking into it currently but locally every single integ is failing for me, probably something I missed in my set up with the version bump but if not, I'll open an issue in FTRepo.

amitgalitz:
@jackiehanyang could the failure on windows-ci PR be related to this not being merged yet?

ohltyler:
@jackiehanyang can we merge this? Sounds like CI failures are going to be fixed with #330?

jackiehanyang:
> @jackiehanyang can we merge this? Sounds like CI failures are going to be fixed with #330?

no I don't think the CI failure will be fixed with https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/330, the CI failure looks like related to 3.0 specifically. Also let's 2.4 windows pr merged before this since it has a higher priority

jackiehanyang:
looked into the error message for failed remote integ tests workflow. It's failing due to flaky integ tests, nothing related to 3.0 specific issue. So it's good to merge @ohltyler

jackiehanyang (approved):

ohltyler:
@jackiehanyang fyi, because I linked the issue, it would have closed automatically when merging :)

PR #346, opened 2022-11-29
Add branch constants in CI workflow for 2.x

ohltyler:
Signed-off-by: Tyler Ohlsen <ohltyler@amazon.com>

### Description

Same as #345 but for 2.x branch.

### Check List

- [x] Commits are signed per the DCO using `--signoff`

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #395, opened 2023-01-10
Add 2.5 release notes

ohltyler:

### Description

Add 2.5 release notes

### Check List

### DIFF
--- a/release-notes/opensearch-anomaly-detection-dashboards.release-notes-2.5.0.0.md
+++ b/release-notes/opensearch-anomaly-detection-dashboards.release-notes-2.5.0.0.md
@@ -1,0 +1,20 @@
+## Version 2.5.0.0 Release Notes
+
+Compatible with OpenSearch Dashboards 2.5.0
+
+### Bug Fixes
+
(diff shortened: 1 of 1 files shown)


amitgalitz:
nit: https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/346 was only merged to 2.x and didn't show up in the release drafter cause of that. On a related note, we should probably change our release drafter to run on both main and 2.x if that's possible to catch things in 2.x release that aren't in main. In this case its just a version bump so not significant but there could be other things

ohltyler:
> nit: #346 was only merged to 2.x and didn't show up in the release drafter cause of that. On a related note, we should probably change our release drafter to run on both main and 2.x if that's possible to catch things in 2.x release that aren't in main. In this case its just a version bump so not significant but there could be other things

It's possible but it gets messy because all of the backport PRs would show up in the draft, so the majority of the time there would be duplicate entries for everything. I'll add #346 though.

kaituo (approved):

amitgalitz (approved):

PR #448, opened 2023-03-29
Adding expression function to retrieve anomalies 

amitgalitz:
### Description

Creates and registers an expression function to retrieve anomalies from AD as a vis-layer to be rendered on top of visualizations in OSD.

The expression function takes a detectorId argument and utilizes the date range in the context to fetch anomalies. The anomalies are then returned as vis-layer which visualize_embeddable in OSD renders onto the given visualization.

This PR is rebased directly from the featureAnywhere project. It doesn't have the changes to public/plugins.tsx as that code is yet unmerged. The only lines added there are: 

```
plugins.expressions.registerFunction(overlayAnomaliesFunction);
setClient(core.http);
```

### Check List

### DIFF
--- a/opensearch_dashboards.json
+++ b/opensearch_dashboards.json
@@ -3,12 +3,12 @@
   "version": "3.0.0.0",
   "opensearchDashboardsVersion": "3.0.0",
   "configPath": ["anomaly_detection_dashboards"],
-  "requiredPlugins": ["navigation", 
+  "requiredPlugins": [
     "opensearchDashboardsUtils",
     "expressions",
-    "data"
+    "data",
+    "visAugmenter"
   ],
-  "optionalPlugins":[],  
   "server": true,
   "ui": true
 }
--- a/opensearch_dashboards.json
+++ b/opensearch_dashboards.json
@@ -3,8 +3,12 @@
   "version": "3.0.0.0",
   "opensearchDashboardsVersion": "3.0.0",
   "configPath": ["anomaly_detection_dashboards"],
-  "requiredPlugins": ["navigation"],
-  "optionalPlugins": [],
+  "requiredPlugins": ["navigation", 
+    "opensearchDashboardsUtils",
+    "expressions",
+    "data"
+  ],
+  "optionalPlugins":[],  
   "server": true,
   "ui": true
 }
--- a/public/expressions/overlay_anomalies.ts
+++ b/public/expressions/overlay_anomalies.ts
@@ -1,0 +1,193 @@
+// /*
+//  * SPDX-License-Identifier: Apache-2.0
+//  *
+//  * The OpenSearch Contributors require contributions made to
+//  * this file be licensed under the Apache-2.0 license or a
+//  * compatible open source license.
+//  *
+//  * Any modifications Copyright OpenSearch Contributors. See
+//  * GitHub history for details.
+//  */
+
+import { get } from 'lodash';
+import { i18n } from '@osd/i18n';
+import { ExpressionFunctionDefinition } from '../../../../src/plugins/expressions/public';
+import {
+  VisLayerTypes,
+  VisLayers,
+  ExprVisLayers,
+  PluginResource,
+} from '../../../../src/plugins/vis_augmenter/public';
+import {
+  TimeRange,
+  calculateBounds,
+} from '../../../../src/plugins/data/common';
+import {
+  getAnomalySummaryQuery,
+  parsePureAnomalies,
+} from '../pages/utils/anomalyResultUtils';
+import { AD_NODE_API } from '../../utils/constants';
+import { AnomalyData } from '../models/interfaces';
+import { getClient } from '../services';
+import {
+  PointInTimeEventsVisLayer,
+  VisLayerError,
+  VisLayerErrorTypes,
+} from '../../../../src/plugins/vis_augmenter/public';
+import { PLUGIN_NAME } from '../../public/utils/constants';
+import { NO_PERMISSIONS_KEY_WORD } from '../../server/utils/helpers';
+
+type Input = ExprVisLayers;
+type Output = Promise<ExprVisLayers>;
+
+interface Arguments {
+  detectorId: string;
+}
+
+const name = 'overlay_anomalies';
+
+export type OverlayAnomaliesExpressionFunctionDefinition =
+  ExpressionFunctionDefinition<'overlay_anomalies', Input, Arguments, Output>;
+
+// This gets all the needed anomalies for the given detector ID and time range
+const getAnomalies = async (
+  detectorId: string,
+  startTime: number,
+  endTime: number
+): Promise<AnomalyData[]> => {
+  const anomalySummaryQuery = getAnomalySummaryQuery(
+    startTime,
+    endTime,
+    detectorId,
+    undefined,
+    false
+  );
+
+  // We set the http client in the plugin.ts setup() fn. We pull it in here to make a
+  // server-side call directly.
+  // Note we can't use the redux fns here (e.g., searchResults()) since it requires
+  // hooks (e.g., useDispatch()) which doesn't make sense in this context, plus is not allowed by React.
+  const anomalySummaryResponse = await getClient().post(
+    `..${AD_NODE_API.DETECTOR}/results/_search`,
+    {
+      body: JSON.stringify(anomalySummaryQuery),
+    }
+  );
+
+  return parsePureAnomalies(anomalySummaryResponse);
+};
+
+// This takes anomalies and returns them as vis layer of type PointInTimeEvents
+const convertAnomaliesToLayer = (
+  anomalies: AnomalyData[],
+  ADPluginResource: PluginResource
+): PointInTimeEventsVisLayer => {
+  const events = anomalies.map((anomaly: AnomalyData) => {
+    return {
+      timestamp: anomaly.startTime + (anomaly.endTime - anomaly.startTime) / 2,
+      metadata: {},
+    };
+  });
+  return {
+    originPlugin: 'anomaly-detection',
+    type: VisLayerTypes.PointInTimeEvents,
+    pluginResource: ADPluginResource,
+    events: events,
+  } as PointInTimeEventsVisLayer;
+};
+
+export const overlayAnomaliesFunction =
+  (): OverlayAnomaliesExpressionFunctionDefinition => ({
+    name,
+    type: 'vis_layers',
+    inputTypes: ['vis_layers'],
+    help: i18n.translate('data.functions.overlay_anomalies.help', {
+      defaultMessage: 'Add an anomaly vis layer',
+    }),
+    args: {
+      detectorId: {
+        types: ['string'],
+        default: '""',
+        help: '',
+      },
+    },
+
+    async fn(input, args, context): Promise<ExprVisLayers> {
+      // Parsing all of the args & input
+      const detectorId = get(args, 'detectorId', '');
+      const timeRange = get(
+        context,
+        'searchContext.timeRange',
+        ''
+      ) as TimeRange;
+      const origVisLayers = get(input, 'layers', [] as VisLayers) as VisLayers;
+      const parsedTimeRange = timeRange ? calculateBounds(timeRange) : null;
+      const startTimeInMillis = parsedTimeRange?.min?.unix()
+        ? parsedTimeRange?.min?.unix() * 1000
+        : undefined;
+      const endTimeInMillis = parsedTimeRange?.max?.unix()
+        ? parsedTimeRange?.max?.unix() * 1000
+        : undefined;
+      const ADPluginResource = {
+        type: 'anomaly-detection-type',
+        id: detectorId,
+        name: PLUGIN_NAME,
+        urlPath: `${PLUGIN_NAME}#/detectors`,
+      };
+      try {
+        if (startTimeInMillis === undefined || endTimeInMillis === undefined) {
+          throw new RangeError('start or end time invalid');
+        }
+        const anomalies: AnomalyData[] = await getAnomalies(
+          detectorId,
+          startTimeInMillis,
+          endTimeInMillis
+        );
+        const anomalyLayer = convertAnomaliesToLayer(
+          anomalies,
+          ADPluginResource
+        );
+        return {
+          type: 'vis_layers',
+          layers: origVisLayers
+            ? origVisLayers.concat(anomalyLayer)
+            : ([anomalyLayer] as VisLayers),
+        };
+      } catch (error) {
+        console.log('Anomaly Detector - Unable to get anomalies: ', error);
+        let visLayerError: VisLayerError = {} as VisLayerError;
+        if (
+          typeof error === 'string' &&
+          error.includes(NO_PERMISSIONS_KEY_WORD)
+        ) {
+          visLayerError = {
+            type: VisLayerErrorTypes.PERMISSIONS_FAILURE,
+            message: error, // might just change this to a generic message like rest of AD plugin
+          };
+        } else {
+          visLayerError = {
+            type: VisLayerErrorTypes.FETCH_FAILURE,
+            message:
+              error === 'string'
+                ? error
+                : error instanceof Error
+                ? error.message
+                : '',
+          };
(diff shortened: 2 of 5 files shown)


opensearch_dashboards.json
```diff
@@ -3,8 +3,12 @@
   "version": "3.0.0.0",
   "opensearchDashboardsVersion": "3.0.0",
   "configPath": ["anomaly_detection_dashboards"],
-  "requiredPlugins": ["navigation"],
-  "optionalPlugins": [],
+  "requiredPlugins": ["navigation",
```
ohltyler, line 6: nit / unrelated: while you're changing this file, I think we can go ahead and remove `navigation`. It was set when the plugin was initially created before this was understood and it isn't needed.

public/expressions/overlay_anomalies.ts
```diff
@@ -0,0 +1,193 @@
+// /*
+//  * SPDX-License-Identifier: Apache-2.0
+//  *
+//  * The OpenSearch Contributors require contributions made to
+//  * this file be licensed under the Apache-2.0 license or a
+//  * compatible open source license.
+//  *
+//  * Any modifications Copyright OpenSearch Contributors. See
+//  * GitHub history for details.
+//  */
```
ohltyler, line 10: For new files we can use the short form:
```
/*
 * Copyright OpenSearch Contributors
 * SPDX-License-Identifier: Apache-2.0
 */
```

public/expressions/overlay_anomalies.ts
```diff
@@ -0,0 +1,193 @@
...
+} from '../pages/utils/anomalyResultUtils';
+import { AD_NODE_API } from '../../utils/constants';
+import { AnomalyData } from '../models/interfaces';
+import { getClient } from '../services';
+import {
+  PointInTimeEventsVisLayer,
+  VisLayerError,
+  VisLayerErrorTypes,
+} from '../../../../src/plugins/vis_augmenter/public';
+import { PLUGIN_NAME } from '../../public/utils/constants';
```
ohltyler, line 37: nit: can be simplified to `'../utils/constants'`

public/services.ts
```diff
@@ -0,0 +1,20 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Any modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
```
ohltyler, line 10: Same as https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/448/files#r1153574259

ohltyler:
This PR is rebased directly from the featureAnywhere project. It doesn't have the changes to public/plugins.tsx as that code is yet unmerged.

Any reason these changes can't be part of this PR? Otherwise after merging, the branch won't be able to use the expr fn since it isn't registered and `client` service isn't set.

opensearch_dashboards.json
```diff
@@ -3,8 +3,12 @@
...
   "opensearchDashboardsVersion": "3.0.0",
   "configPath": ["anomaly_detection_dashboards"],
-  "requiredPlugins": ["navigation"],
-  "optionalPlugins": [],
+  "requiredPlugins": ["navigation", 
+    "opensearchDashboardsUtils",
+    "expressions",
+    "data"
+  ],
+  "optionalPlugins":[],
```
ohltyler, line 11: this can be removed

public/expressions/overlay_anomalies.ts
```diff
@@ -0,0 +1,193 @@
...
+    endTime,
+    detectorId,
+    undefined,
+    false
+  );
+
+  // We set the http client in the plugin.ts setup() fn. We pull it in here to make a
+  // server-side call directly.
+  // Note we can't use the redux fns here (e.g., searchResults()) since it requires
+  // hooks (e.g., useDispatch()) which doesn't make sense in this context, plus is not allowed by React.
```
ohltyler, line 69: I think we can remove this comment. I had added it initially when trying to develop it but it doesn't seem to add much value now

public/expressions/overlay_anomalies.ts
```diff
@@ -0,0 +1,193 @@
...
+  ADPluginResource: PluginResource
+): PointInTimeEventsVisLayer => {
+  const events = anomalies.map((anomaly: AnomalyData) => {
+    return {
+      timestamp: anomaly.startTime + (anomaly.endTime - anomaly.startTime) / 2,
+      metadata: {},
+    };
+  });
+  return {
+    originPlugin: 'anomaly-detection',
```
ohltyler, line 92: I wonder if we should extract this into a constant. It might make it easier to track when processing the values on OSD side.

public/expressions/overlay_anomalies.ts
```diff
@@ -0,0 +1,193 @@
...
+        : undefined;
+      const endTimeInMillis = parsedTimeRange?.max?.unix()
+        ? parsedTimeRange?.max?.unix() * 1000
+        : undefined;
+      const ADPluginResource = {
+        type: 'anomaly-detection-type',
+        id: detectorId,
+        name: PLUGIN_NAME,
+        urlPath: `${PLUGIN_NAME}#/detectors`,
+      };
```
ohltyler, line 136: A few notes on this part:
- `type` will be the header used when categorizing and grouping the `VisLayer`s on the view events flyout, so ideally it should be something like `Anomaly Detectors` (this could be pulled into a constant as well)
- `urlPath` should be the full path to the relevant details page for the plugin resource - in our case, `${PLUGIN_NAME}#/detectors/${detectorId}/results` probably

I'll follow up by adding some more descriptions on these fields in the interface definitions. It might be worth adding some comments wherever these are implemented here too to prevent confusion.

public/expressions/overlay_anomalies.ts
amitgalitz: Here also I wasn't 100% sure if it should be `anomaly-detection` or `anomaly-detection-dashboards` or even `anomalyDetectionDashboards` as we list on opensearch_dashboards.json. Definitely agree it should be a constant

public/expressions/overlay_anomalies.ts
```diff
@@ -0,0 +1,193 @@
...
+      try {
+        if (startTimeInMillis === undefined || endTimeInMillis === undefined) {
+          throw new RangeError('start or end time invalid');
+        }
+        const anomalies: AnomalyData[] = await getAnomalies(
```
ohltyler, line 141: nit: duplicate type declaration not needed since it's defined in `getAnomalies()`

public/expressions/overlay_anomalies.ts
```diff
@@ -0,0 +1,193 @@
...
+      } catch (error) {
+        console.log('Anomaly Detector - Unable to get anomalies: ', error);
+        let visLayerError: VisLayerError = {} as VisLayerError;
+        if (
+          typeof error === 'string' &&
+          error.includes(NO_PERMISSIONS_KEY_WORD)
+        ) {
+          visLayerError = {
+            type: VisLayerErrorTypes.PERMISSIONS_FAILURE,
+            message: error, // might just change this to a generic message like rest of AD plugin
```
ohltyler, line 165: can you add `TODO` to this comment so it can be more easily tracked?

public/expressions/overlay_anomalies.ts
```diff
@@ -0,0 +1,193 @@
...
+  });
+  return {
+    originPlugin: 'anomaly-detection',
+    type: VisLayerTypes.PointInTimeEvents,
+    pluginResource: ADPluginResource,
+    events: events,
+  } as PointInTimeEventsVisLayer;
+};
+
+export const overlayAnomaliesFunction =
```
ohltyler, line 99: Just for further documentation, could you add a comment for this function describing more about its purpose / how it's adding a vislayer onto an existing array of vislayers? Making that more clear can help us reference it when adding more documentation / tutorials on how other plugins may integrate.

public/services.ts
ohltyler, line 1: nice! thanks for creating this file. Makes it cleaner for having a clear defined set of the dependencies being used from the required plugins.

public/expressions/overlay_anomalies.ts
ohltyler: The main purpose of this field is to provide another way to categorize the `VisLayer`s. Few advantages include:
1. Plugins having the opportunity to filter all `VisLayer`s that match the plugin's `originPlugin` field for a particular embeddable to post-process the data, instead of having to re-fetch saved objects (not sure if applicable given the current way plugins are consuming visualize_embeddable, but it leaves that option open)
2. More categorizing/filtering within the view events page or other OSD-owned components that are displaying the `VisLayer`s in some organized manner

Perhaps these `originPlugin` constants, and `type` constants can be put in a dedicated constants file specific to `VisLayer`s / this feature, so we can kind of split up constants that are only used within our plugin vs. constants that may be utilized in OSD via the render framework.

ohltyler: To clarify, this field isn't intended to be user-facing so it can be something internal like `anomalyDetectionDashboards` (this one makes sense to me to be consistent with the internal plugin name)

public/expressions/overlay_anomalies.ts
ohltyler: It's difficult to formalize this across plugins without limiting flexibility, since all we care about here is a string. I think we can leave this up to the plugins for now, as well as document the use of it in OSD. If that's still causing problems down the road we could revisit the interfaces.

public/plugin.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { PLUGIN_NAME } from './utils/constants';
+import { getActions } from './utils/contextMenu/action';
+import { setClient, setSavedFeatureAnywhereLoader } from './services';
+import { overlayAnomaliesFunction } from './expressions';
+
+declare module '../../../src/plugins/ui_actions/public' {
+  export interface ActionContextMapping {
+    [ACTION_AD]: {};
+  }
+}
```
ohltyler, line 29: looks like this is copied over from a separate PR, can these action-related code changes be omitted from this PR? They can be added to the file as part of that PR so the history is clean.

ohltyler: fixed in latest commit, thanks.

public/expressions/overlay_anomalies.ts
amitgalitz: will change

opensearch_dashboards.json
amitgalitz: removed, also added visAugmenter which I forgot to add originally here when simplifying the PR

public/expressions/overlay_anomalies.ts
```diff
@@ -0,0 +1,193 @@
...
+    {
+      body: JSON.stringify(anomalySummaryQuery),
+    }
+  );
+
+  return parsePureAnomalies(anomalySummaryResponse);
+};
+
+// This takes anomalies and returns them as vis layer of type PointInTimeEvents
+const convertAnomaliesToLayer = (
```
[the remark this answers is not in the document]
amitgalitz: convertAnomaliesToPointInTimeEventsVisLayer makes sense to me as its the most accurate to what method is doing, good call.

public/expressions/overlay_anomalies.ts
amitgalitz: I am all for documenting as I understand the limitation around formalizing a specific standard. Good call adding some description in the interface definition. I will add some comment in this code and potentially we can also add more info in documentation for future plugin onboarding

public/expressions/overlay_anomalies.ts
amitgalitz: I guess though, do you want the originPlugin constants here for the `VisLayer` interface and the name for the `PluginResource` be exactly the same?

public/expressions/overlay_anomalies.ts
ohltyler: Forgot to also add:
`name` should be the name of the plugin resource, not the name of the plugin name. In this case, detector name

public/expressions/overlay_anomalies.ts
ohltyler: >  I guess this could be good to point out either in comments above the interface or to add to a plugin onboarding documentation of some sort

Yeah, that will be captured as part of https://github.com/opensearch-project/OpenSearch-Dashboards/issues/3383

public/expressions/overlay_anomalies.ts
amitgalitz: Added

public/expressions/overlay_anomalies.ts
amitgalitz: added this

public/expressions/overlay_anomalies.ts
```diff
@@ -77,8 +72,18 @@ const getAnomalies = async (
   return parsePureAnomalies(anomalySummaryResponse);
 };
 
+const getDetectorName = async (detectorId: string) => {
+  const resp = await getClient().get(`..${AD_NODE_API.DETECTOR}/${detectorId}`);
```
ohltyler, line 76: Just to confirm, if there is failures coming from the cluster will this throw an error, or just store the error in `resp`? I Imagine we want to make sure an error is thrown so its caught in the `catch` block you have to process and produce an error `VisLayer`

amitgalitz: good catch, I actually fixed this by moving the call to getDetectorName inside the try, catch. Now if there is an error from the cluster it is automatically caught in the catch (like  status 4xx). If we get some response but no detector name, I manually throw the error which gets translated to a fetching failure.

ohltyler: It still seems a little hacky to be basing success/failure on if the name is empty or not. In fact, can we just use existing redux functions here to handle the errors for us? For example, [getDetector](https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/public/redux/reducers/ad.ts#L391)? We can then process any errors seen, similar to [this example](https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/public/pages/ReviewAndCreate/containers/ReviewAndCreate.tsx#L184-L199) with `then()`/`catch()` blocks. Lmk what you think.

ohltyler: Similar thought regarding the `anomalySummaryResponse` on L65. Maybe we can use the [searchResults](https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/public/redux/reducers/anomalyResults.ts#L124) redux fn.

ohltyler: I think the original intention of using `client()` standalone here was if other cluster calls needed to happen. But given that we only need standard ones we already have that are integrated with redux's already-consumed HTTP client, I think we're ok.

amitgalitz: I have tried using the dispatch here in a few several ways and I keep coming back to the fact that we need to wrap the class in a provider in order to use our redux methods. The error I get here is: `Uncaught Error: could not find react-redux context value; please ensure the component is wrapped in a <Provider>`. All our other uses of redux is within the parent component `<Main>` which is wrapped by the redux provider.

amitgalitz: Do you see other ways we could optimize this to make you more confident in just using the client? I believe the current catching mechanism works as if there is an actual error from the client like a network issue it gets caught in the catch and I have tested this (with mistyping the API for example which gave an error) and if there is some other reason the name can't be returned (I doubt this will actually happen) then that also gets caught.

ohltyler: Great - I think that's fine to just use the client here. Given that this is a unique context not wrapped in the provider, I think that's ok.

As long as we have sufficient error catching & handling I'm ok with this

ohltyler (approved):

jackiehanyang (approved):

jackiehanyang:
some unit tests are failing, other than that, lgtm

amitgalitz:
Unit tests only failing on macos and it passed locally. Not able to re-run tests on this PR.

PR #442, opened 2023-03-20
Associated detectors flyout

amitgalitz:
### Description

This change adds the associated detectors flyout as part of the feature-anywhere project. This flyout allows you to view all the detectors associated with a particular visualization. From the flyout a user can unlink a detector from a visualization or search up their associated detectors and click on it to view the detector details on the AD plugin. 

This PR also includes code from an unmerged PRs. All my specific changes are starting from commit: https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/442/commits/60d63ece319d5294b3e41d1220cb56dc11cdeffc

Additionally in this PR the "Associate Detector" button doesn't do anything, the reason for this is that it relies on some other changes in the CreateAnomalyDetector flyout that aren't merged yet. I will add this functionality on another PR.

Screenshot of flyout

Screenshot 2023-03-27 at 10 00 21 AM

### Check List
- [x] Commits are signed per the DCO using `--signoff`

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #449, opened 2023-03-30
Associated detectors flyout through UI actions

amitgalitz:
### Description

This change adds the associated detectors flyout as part of the feature-anywhere project. This flyout allows you to view all the detectors associated with a particular visualization. From the flyout a user can unlink a detector from a visualization or search up their associated detectors and click on it to view the detector details on the AD plugin.

In this PR the "Associate Detector" button doesn't do anything, the reason for this is that it relies on some other changes in the CreateAnomalyDetector flyout that aren't merged yet. I will add this functionality on another PR.

This code also might have some build failures because it relies on files such as [public/action/ad_dashboard_action.tsx](https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/442/files#diff-973cb51dfb9b0c8247635668d26f0988132cb593eba59e2aa7983c47736bd7c4R2) to run correctly. The reason this file as well as some of the CreateAnomalyDetector files aren't included is because it will be merged first by Jackie's PR. 

This PR also requires additional testing that I will be adding. 

Screenshot of flyout:
Screenshot 2023-03-27 at 10 00 21 AM

### Check List


public/components/FeatureAnywhereContextMenu/AssociatedDetectors/components/ConfirmUnlinkDetectorModal/ConfirmUnlinkDetectorModal.tsx
```diff
@@ -0,0 +1,80 @@
...
+  return (
+    <EuiOverlayMask>
+      <EuiModal
+        data-test-subj="startDetectorsModal"
+        onClose={props.onHide}
+        maxWidth={450}
+      >
+        <EuiModalHeader>
+          <EuiModalHeaderTitle>
+            {'Remove association?'}&nbsp;
```
ohltyler, line 43: Whats the reason for the extra space (`&nbsp;`) here?

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/components/ConfirmUnlinkDetectorModal/ConfirmUnlinkDetectorModal.tsx
```diff
@@ -0,0 +1,80 @@
...
+          </EuiModalHeaderTitle>
+        </EuiModalHeader>
+        <EuiModalBody>
+          <EuiText>
+            Removing association unlinks {props.detector.name} detector from the
```
ohltyler, line 48: Maybe this is from UX, but this whole sentence seems grammatically weird. I would prefer something like
`Removing the association unlinks detector '<detector-name>' from the visualization, but does not delete it`

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/components/ConfirmUnlinkDetectorModal/ConfirmUnlinkDetectorModal.tsx
```diff
@@ -0,0 +1,80 @@
...
+        </EuiModalBody>
+        <EuiModalFooter>
+          {isLoading ? null : (
+            <EuiButtonEmpty
+              data-test-subj="cancelButton"
+              onClick={props.onHide}
+            >
+              Cancel
+            </EuiButtonEmpty>
+          )}
```
ohltyler, line 62: nit - could we change the loading state to just set `isLoading={true}` for the button, instead of having no button at all? In other words, remove the conditional and just populate the button's `isLoading` prop with the value of the existing `isLoading` prop that you are passing in.

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/containers/AssociatedDetectors.tsx
```diff
@@ -0,0 +1,347 @@
...
+import { getDetectorList } from '../../../../redux/reducers/ad';
+import {
+  prettifyErrorMessage,
+  NO_PERMISSIONS_KEY_WORD,
+} from '../../../../../server/utils/helpers';
+import { SavedObjectLoader } from '../../../../../../../src/plugins/saved_objects/public';
+import { EmptyAssociatedDetectorFlyoutMessage } from '../components/EmptyMessage/EmptyMessage';
+import { ISavedAugmentVis } from '../../../../../../../src/plugins/vis_augmenter/public';
+import { ASSOCIATED_DETECTOR_ACTION } from '../utils/constants';
+import { ConfirmUnlinkDetectorModal } from '../components/ConfirmUnlinkDetectorModal/ConfirmUnlinkDetectorModal';
```
ohltyler, line 38: nit - you can add `index.ts` files for each directory and sub-directory to handle the exports. This way it simplifies imports since you just have to reference directories instead of the implemented file. e.g., changing the above to 
```
import { ConfirmUnlinkDetectorModal } from '../components';
```

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/containers/AssociatedDetectors.tsx
```diff
@@ -0,0 +1,347 @@
...
+
+  useEffect(() => {
+    getDetectors();
+  }, []);
+
+  // Handles all changes in the assoicated detectors such as unlinking or new detectors associated
+  useEffect(() => {
+    // Gets all augmented saved objects
+    savedObjectLoader
+      .findAll()
```
ohltyler, line 125: You should be able to pass some arguments to `findAll()` to handle filtering for you ([ref](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/feature/feature-anywhere/src/plugins/saved_objects/public/saved_object/saved_object_loader.ts#L131)). For example, passing in search fields on `originPlugin` and/or `visId` with the values you want. This way you wouldn't need to have all of the helper fns.

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/containers/AssociatedDetectors.tsx
```diff
@@ -0,0 +1,347 @@
...
+            get(savedObject, 'pluginResourceId', '') === detectorToUnlink.id
+        );
+        const savedObjectToUnlinkId = get(savedAugmentToUnlink, 'id', '');
+        await savedObjectLoader
+          .delete(savedObjectToUnlinkId)
+          .then(async (resp: any) => {
+            core.notifications.toasts.addSuccess({
+              title: `Association removed between the ${detectorToUnlink.name}
+              and the ${embeddableTitle} visualization`,
+              text: "The detector's anomalies do not appear on the visualization. Refresh your dashboard to update the visualization",
```
ohltyler, line 198: 2 things:
1. Should the first sentence be something like "The detector's anomalies may still appear on the visualization"? Because from the second sentence, it seems that's the case
2. Was it determined from UX that we don't want to auto-refresh for the user?

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/containers/AssociatedDetectors.tsx
```diff
@@ -0,0 +1,347 @@
...
+  };
+
+  // TODO: this part is incomplete because it is pending on a different PR that will have all the associate existing changes
+  const onAssociateExistingDetector = async () => {
+    console.log('inside create anomaly detector');
+  };
+
+  const handleUnlinkDetectorAction = (detector: DetectorListItem) => {
+    setDetectorToUnlink(detector);
+    if (!isEmpty(detector)) {
```
ohltyler, line 254: When would the stored detector list item be empty?

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/containers/AssociatedDetectors.tsx
```diff
@@ -0,0 +1,347 @@
...
+      <EuiFlyout ownFocus size="m" paddingSize="m" onClose={closeFlyout}>
+        <EuiFlyoutHeader hasBorder>
+          <EuiTitle size="s">
+            <h2 id="associated-detectors__title">
+              Associated anomaly detectors
+            </h2>
+          </EuiTitle>
+        </EuiFlyoutHeader>
+        <EuiFlyoutBody style={{ overflowY: 'auto' }}>
+          {getUnlinkConfirmModal()}
```
ohltyler, line 320: nit - would it be cleaner to just instantiate the confirm modal jsx here instead of embedding within a function to fetch it?

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/containers/AssociatedDetectors.tsx
```diff
@@ -0,0 +1,347 @@
...
+            <EuiFlexItem>
+              <EuiButton
+                fill
+                iconType="link"
+                onClick={() => {
+                  onAssociateExistingDetector();
+                }}
+              >
+                Associate a detector
+              </EuiButton>
```
ohltyler, line 336: Should we add an `isLoading` prop to the button here so that users cant click on it if there is still data being loaded?

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/styles.scss
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+@import '@elastic/eui/src/global_styling/variables/index';
```
ohltyler, line 6: Can this import be removed?

public/components/FeatureAnywhereContextMenu/Container/Container.tsx
```diff
@@ -0,0 +1,29 @@
+import React, { useState } from 'react';
+import './styles.scss';
+import AssociatedDetectors from '../AssociatedDetectors/containers/AssociatedDetectors';
+
+const Container = ({ startingFlyout, ...props }) => {
```
ohltyler, line 5: nit - can we have this named something else to be more specific here? `Container` is a bit vague.

public/components/FeatureAnywhereContextMenu/Container/Container.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const { embeddable } = props;
+  const index = [{ label: embeddable?.vis?.data?.indexPattern?.title }];
```
ohltyler, line 7: we should handle `index` better such that if it doesn't exist, it can be passed down to `Flyout` component with that information. At least specifying the type as `string | undefined` can be helpful here

public/components/FeatureAnywhereContextMenu/Container/styles.scss
```diff
@@ -0,0 +1,7 @@
+.context-menu {
+  &__flyout {
+    &.euiFlyout--medium {
+      width: 740px;
```
ohltyler, line 4: I believe this is copied over from here: https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/450#discussion_r1154971204

But pointing out my concerns with that here as well

public/utils/contextMenu/styles.scss
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+@import '@elastic/eui/src/global_styling/variables/index';
```
ohltyler, line 6: what is the purpose of this import?

public/components/FeatureAnywhereContextMenu/Container/styles.scss
amitgalitz: Yes I actually agree, I think we should use small or medium. I missed this while trying to align with Jackie's code. I have the associated detectors style set to medium for the flyout, will check with UX if its okay to go with that for all ad flyouts

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/containers/AssociatedDetectors.tsx
amitgalitz: Hmm I don't think in this case we need an isLoading since this is just to send them to the associate with existing detector flyout. It wont associate a detector so it has no direct dependency to the data being loaded. The only use case I can see is that if user clicks on this cause they see they don't have any detectors currently associated with the viz but that would have a loading component there. Users might want to still go to associate an existing detector flyout while all the current ones are loading. I can bring this up with UX team though.

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/components/ConfirmUnlinkDetectorModal/ConfirmUnlinkDetectorModal.tsx
amitgalitz: yes the wording here was given by UX. I see your point, I will bring it up with UX/techinical writers and fix up the wording here.

public/components/FeatureAnywhereContextMenu/Container/Container.tsx
amitgalitz: Looking into this, not as easy as `string | undefined` as right now it returns '{ label: any; }[]'

public/components/FeatureAnywhereContextMenu/Container/Container.tsx
amitgalitz: Does `AnywhereParentFlyout` sounds good?

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/containers/AssociatedDetectors.tsx
amitgalitz: Would you prefer that this is used then what I already have coded, just in terms I know this already works and has been tested. We can also have any future optimization or changes in our control here.

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/containers/AssociatedDetectors.tsx
amitgalitz: 1. This makes sense to me, I'll confirm with UX as they gave wording but it does seem like a logic error.
2. Yes currently UX is okay with not auto-refreshing, also both alerting folks and I haven't been able to get the auto refresh of the dashboard to work even if we wanted it.

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/containers/AssociatedDetectors.tsx
amitgalitz: I actually don't think it would be possible on a second thought here since its coming directly from the list of detectors. I think originally I was thinking that there would be some lag if detector was already unlinked but I guess it wouldn't be in this operation cause its coming from the stored list and could maybe happen on the crud operations where i do have error checking

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/containers/AssociatedDetectors.tsx
amitgalitz: I can do this, just wanted to a shorted component here but no strong preference

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/components/ConfirmUnlinkDetectorModal/ConfirmUnlinkDetectorModal.tsx
ohltyler: Sounds good. Not a blocker to me - would be great to follow up.

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/components/ConfirmUnlinkDetectorModal/ConfirmUnlinkDetectorModal.tsx
ohltyler: Got it - my initial thought was this would be in loading state when opening it, instead of closing/finishing the action. Makes sense to me to leave as-is.

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/containers/AssociatedDetectors.tsx
ohltyler: I would prefer the way I recommended since it would allow us to leverage OpenSearch and be more optimized and simpler. Also, it would be less code we have to maintain or test.

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/containers/AssociatedDetectors.tsx
ohltyler: Got it.

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/containers/AssociatedDetectors.tsx
ohltyler: Cool - let's go ahead and remove it

public/components/FeatureAnywhereContextMenu/Container/Container.tsx
ohltyler: fine with me :)

public/components/FeatureAnywhereContextMenu/Container/Container.tsx
ohltyler: Right - or perhaps `{ label: string | undefined }`, or set default as empty string:
```
const index = [{ label: get(embeddable, 'vis.data.indexPattern.title', '') }];
```

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/containers/AssociatedDetectors.tsx
amitgalitz: Discussed offline, filtering through FindAll is limited to title and description, issue is here: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/3566

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/containers/AssociatedDetectors.tsx
amitgalitz: I added a few index.ts to handle exports, should we do this even for every components folder that just has one file? didn't want to cause way too many files.

public/utils/contextMenu/getActions.tsx
```diff
@@ -74,6 +75,18 @@ export const getActions = ({ core, plugins }) => {
...
       onClick: getOnClick('associated'),
     },
+    {
+      id: 'documentationAnomalyDetector',
+      title: <DocumentationTitle />,
+      icon: 'documentation' as EuiIconType,
+      order: 98,
+      onExecute: () => {
+        window.open(
+          'https://opensearch.org/docs/latest/monitoring-plugins/anomaly-detection/index/',
```
ohltyler, line 85: Let's leverage BASE_DOCS_LINK here: https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/public/utils/constants.ts#L54

ohltyler: And if the docs link needs to be updated, can you help update?

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/containers/AssociatedDetectors.tsx
amitgalitz: Added this

public/utils/contextMenu/getActions.tsx
amitgalitz: Changed to use the doc link constant. No need to update here actually.

ohltyler:
LGTM! Will wait to fully approve when rebased after #482 is merged first.

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/containers/AssociatedDetectors.tsx
```diff
@@ -0,0 +1,337 @@
...
+            (savedObj) => get(savedObj, 'visId', '') === embeddable.vis.id
+          );
+
+        // find saved Augment object matching detector we want to unlink
+        // There should only be one detector and vis pairing
+        const savedAugmentToUnlink = savedAugmentForThisVisualization.find(
+          (savedObject) =>
+            get(savedObject, 'pluginResourceId', '') === detectorToUnlink.id
+        );
+        const savedObjectToUnlinkId = get(savedAugmentToUnlink, 'id', '');
```
ohltyler, line 193: nit: you could combine all of these into a chained set of cmds to simplify it and to remove having to keep intermediate consts.

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/containers/AssociatedDetectors.tsx
```diff
@@ -0,0 +1,337 @@
...
+    // Filter all savedAugmentObjects that aren't linked to the specific visualization
+    const savedAugmentForThisVisualization: ISavedAugmentVis[] =
+      savedAugmentObjects.filter(
+        (savedObj) => get(savedObj, 'visId', '') === embeddable.vis.id
+      );
+
+    // Map all detector IDs for all the found augmented vis objects
+    const savedAugmentDetectorsSet = new Set(
+      savedAugmentForThisVisualization.map((savedObject) =>
+        get(savedObject, 'pluginResourceId', '')
```
ohltyler, line 165: `ISavedAugmentVis` interface has changed - see [here](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/4006/files#diff-f3a7084285f53c21fad7d46ef7e11a91926e5568f3ac0c9ef9df96f015f99f2a).

Basically this will mean to change above to:
```
get(savedObject, 'pluginResource.id', '')
```

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/containers/AssociatedDetectors.tsx
```diff
@@ -0,0 +1,337 @@
...
+  };
+
+  const getDetectors = async () => {
+    dispatch(getDetectorList(GET_ALL_DETECTORS_QUERY_PARAMS));
+  };
+
+  // TODO: this part is incomplete because it is pending on a different PR that will have all the associate existing changes
+  const openAssociateDetectorFlyout = async () => {
+    console.log('inside create anomaly detector');
+  };
```
ohltyler, line 238: Is this pending once create detector flyout is merged first?

amitgalitz: and this: https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/484

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/containers/AssociatedDetectors.tsx
```diff
@@ -0,0 +1,337 @@
...
+      ...confirmModalState,
+      isRequestingToClose: true,
+    });
+  };
+
+  const getDetectors = async () => {
+    dispatch(getDetectorList(GET_ALL_DETECTORS_QUERY_PARAMS));
+  };
+
+  // TODO: this part is incomplete because it is pending on a different PR that will have all the associate existing changes
```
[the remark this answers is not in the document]
amitgalitz: No this will be finished with this "associate-existing" PR I commented below

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/containers/AssociatedDetectors.tsx
```diff
@@ -0,0 +1,337 @@
...
+        />
+      );
+    } else {
+      return (
+        <EmptyAssociatedDetectorMessage
+          isFilterApplied={false}
+          embeddableTitle={embeddableTitle}
+        />
+      );
+    }
```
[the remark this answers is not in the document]
amitgalitz: This is to catch if there is anything in the search box, if we are rendering the empty message we need to know if its because there is no selected detectors in the state at all or if its because the user's search result is returning nothing. The message would then differ with something like "There are no detectors matching your search" vs "There are no anomaly detectors associated with ${props.embeddableTitle} visualization."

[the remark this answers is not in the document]
amitgalitz: The renderEmptyMessage() is automatically called by the EuiInMemoryTable when there is no data to display in table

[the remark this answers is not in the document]
ohltyler: Got it, that makes sense

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/containers/AssociatedDetectors.tsx
amitgalitz: I can delete savedAugmentObjects and savedObjectToUnlinkId. Wanted to keep 2 constants for the filter operations to keep them more readable rather then nesting the gets and filters too much.

ohltyler:
Cool, this lgtm. Won't approve until dependent PRs merged

ohltyler (dismissed):

public/components/FeatureAnywhereContextMenu/AnywhereParentFlyout/AnywhereParentFlyout.tsx
```diff
@@ -17,7 +17,7 @@ const AnywhereParentFlyout = ({ startingFlyout, ...props }) => {
   const [selectedDetectorId, setSelectedDetectorId] = useState();
 
   const AnywhereFlyout = {
-    create: AddAnomalyDetector,
```
jackiehanyang, line 20: we shouldn't remove this line, instead, please append the `associated: AssociatedDetectors,` as the second line here

public/components/FeatureAnywhereContextMenu/AnywhereParentFlyout/index.ts
```diff
@@ -0,0 +1,8 @@
+/*
```
jackiehanyang, line 1: this file is not needed, we already have `index.tsx` file.

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/containers/AssociatedDetectors.tsx
```diff
@@ -0,0 +1,335 @@
...
+
+interface ConfirmModalState {
+  isOpen: boolean;
+  action: ASSOCIATED_DETECTOR_ACTION;
+  isListLoading: boolean;
+  isRequestingToClose: boolean;
+  affectedDetector: DetectorListItem;
+}
+
+function AssociatedDetectors({ embeddable, closeFlyout, core, setMode }) {
```
jackiehanyang, line 50: Please avoid passing core as a prop around as tyler suggested in my previous pr - https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/482#discussion_r1198345354

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/containers/AssociatedDetectors.tsx
```diff
@@ -0,0 +1,335 @@
...
+  // Establish savedObjectLoader for all operations on vis_augment saved objects
+  const savedObjectLoader: SavedObjectLoader = getSavedFeatureAnywhereLoader();
+
+  useEffect(() => {
+    if (
+      errorGettingDetectors &&
+      !errorGettingDetectors.includes(SINGLE_DETECTOR_NOT_FOUND_MSG)
+    ) {
+      console.error(errorGettingDetectors);
+      core.notifications.toasts.addDanger(
```
jackiehanyang, line 90: same as above, please remove `core.notifications` usage, instead, just get the notification service to use here

public/components/FeatureAnywhereContextMenu/DocumentationTitle/index.ts
```diff
@@ -0,0 +1,8 @@
+/*
```
jackiehanyang, line 1: this file can be removed

public/utils/constants.ts
```diff
@@ -53,7 +53,8 @@ export const ANOMALY_RESULT_INDEX = '.opendistro-anomaly-results';
 
 export const BASE_DOCS_LINK = 'https://opensearch.org/docs/monitoring-plugins';
 
-export const AD_DOCS_LINK = 'https://opensearch.org/docs/latest/observing-your-data/ad/index/';
+export const AD_DOCS_LINK =
```
jackiehanyang, line 56: you might want to rebase the branch, this constant has already been added - https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/featureAnywhere/public/utils/constants.ts#L56

amitgalitz: I think this is just running prettier on the line

jackiehanyang (approved):

ohltyler (approved):

PR #450, opened 2023-03-30
Register AD as dashboard option

jackiehanyang:
### Description
This pr includes:
- register ad as one of the dashboard options
- setup create ad flyout, intended to left associate existing part empty for @amitgalitz to pick up later
- will send out tests later after the review

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/2916

### Check List

- [ ] New functionality includes testing.
  - [ ] All tests pass
- [ ] New functionality has been documented.
- [ ] Commits are signed per the DCO using `--signoff`

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #482, opened 2023-05-12
Register AD as dashboard context menu option

jackiehanyang:
### Description

This PR includes:
- register AD one of the dashboard context menu option
- register AD actions - `Create anomaly detector`, `Associated anomaly detector`, `Documentation`
- introduce the flyout container - `AnywhereParentFlyout`

Screenshot 2023-05-12 at 2 39 29 PM Screenshot 2023-05-12 at 2 39 02 PM

Excluded the creation flyout context from this pr to keep the pr small. Line 15 in `AnywhereParentFlyout.tsx` file is just a place holder for now.

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/2916

### Check List

  - [ ] All tests pass


public/components/FeatureAnywhereContextMenu/AnywhereParentFlyout/AnywhereParentFlyout.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { useState } from 'react';
+import './styles.scss';
```
amitgalitz, line 2: This can be removed, look at this comment: https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/482/files#r1195437480

public/utils/contextMenu/getActions.tsx
```diff
@@ -0,0 +1,92 @@
...
+      icon: 'gear' as EuiIconType,
+      order: 99,
+      onClick: getOnClick('associated'),
+    },
+    {
+      id: 'documentationAnomalyDetector',
+      title: <DocumentationTitle />,
+      icon: 'documentation' as EuiIconType,
+      order: 98,
+      onExecute: () => {
```
amitgalitz, line 84: We don't define onExecute in the `ad_dashboards_action.tsx` file so this needs to change to `onClick` in order to work

public/utils/constants.ts
```diff
@@ -53,6 +53,8 @@ export const ANOMALY_RESULT_INDEX = '.opendistro-anomaly-results';
 
 export const BASE_DOCS_LINK = 'https://opensearch.org/docs/monitoring-plugins';
 
+export const AD_DOCS_LINK = 'https://opensearch.org/docs/latest/monitoring-plugins/anomaly-detection/index/';
```
amitgalitz, line 56: This link actually was broken for me when testing, I had to use `${BASE_DOCS_LINK}/ad`. LMK if this either works for you or if not then you should try to what worked on my end.

public/plugin.tsx
```diff
@@ -0,0 +1,72 @@
+/*
```
amitgalitz, line 1: I already made some of these changes in my custom expression PR to plugin.ts. Could you just make the necessary changes you need in `plugin.ts` without overwrite my code there. I think you accidently added new plugin.tsx file here instead of using current plugin.ts

public/components/FeatureAnywhereContextMenu/AnywhereParentFlyout/AnywhereParentFlyout.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { useState } from 'react';
```
amitgalitz, line 1: Can you add license to new file:
```
/*
 * Copyright OpenSearch Contributors
 * SPDX-License-Identifier: Apache-2.0
 */
 ```

File: public/action/ad_dashboard_action.tsx [hunk 6]
```diff
@@ -0,0 +1,74 @@
+import { IEmbeddable } from '../../../../src/plugins/dashboard/public/embeddable_plugin';
```
amitgalitz, line 1: similar to below, can you add license to new file

public/utils/contextMenu/getActions.tsx
```diff
@@ -0,0 +1,92 @@
+import React from 'react';
```
amitgalitz, line 1: license here also

public/components/FeatureAnywhereContextMenu/AnywhereParentFlyout/AnywhereParentFlyout.tsx
ohltyler, line 1: Please see original comments: https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/450/files#r1154959694

public/action/ad_dashboard_action.tsx
ohltyler, line 1: Please see original comments: https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/450/files#r1154959694

public/components/FeatureAnywhereContextMenu/AnywhereParentFlyout/AnywhereParentFlyout.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const { embeddable } = props;
+  const indices: { label: string }[] = [
+    { label: get(embeddable, 'vis.data.indexPattern.title', '') },
+  ];
+  const [mode, setMode] = useState(startingFlyout);
+  const [selectedDetectorId, setSelectedDetectorId] = useState();
+
+  const Flyout = {
+    create: AddAnomalyDetector,
+  }[mode];
```
ohltyler, line 16: Can we rename this to something more specific here? `Flyout` seems too generic and could get confused with OUI components

public/components/FeatureAnywhereContextMenu/AnywhereParentFlyout/AnywhereParentFlyout.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+  return (
+    <Flyout
+      {...{
```
ohltyler, line 20: unnecessary spread operator here

public/components/FeatureAnywhereContextMenu/AnywhereParentFlyout/AnywhereParentFlyout.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { useState } from 'react';
+import './styles.scss';
+import { get } from 'lodash';
+import AddAnomalyDetector from '../CreateAnomalyDetector';
+
+const AnywhereParentFlyout = ({ startingFlyout, ...props }) => {
```
ohltyler, line 6: We should make the `props` typesafe here to be consistent with the rest of the plugin, and so it is clear what fields exist in `props`. Given our codebase is mostly all typescript, any javascript-like code copied from alerting should be converted here

public/utils/contextMenu/getActions.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { i18n } from '@osd/i18n';
+import { EuiIconType } from '@elastic/eui/src/components/icon/icon';
```
ohltyler, line 3: path should be able to be shortened to just `'@elastic/eui'`

public/utils/contextMenu/getActions.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { toMountPoint } from '../../../../../src/plugins/opensearch_dashboards_react/public';
+import { Action } from '../../../../../src/plugins/ui_actions/public';
+import { createADAction } from '../../action/ad_dashboard_action';
+import AnywhereParentFlyout from '../../components/FeatureAnywhereContextMenu/AnywhereParentFlyout';
```
ohltyler, line 7: Because we have the `index` file in `components/FeatureAnywhereContextMenu`, you can shorten the import to just be `../../components/FeatureAnywhereContextMenu`

public/utils/contextMenu/getActions.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { Provider } from 'react-redux';
+import { CoreServicesContext } from '../../../public/components/CoreServices/CoreServices';
+import configureStore from '../../redux/configureStore';
+import DocumentationTitle from '../../../public/components/FeatureAnywhereContextMenu/DocumentationTitle/containers/DocumentationTitle';
```
ohltyler, line 11: I don't think this file exists?

public/utils/contextMenu/getActions.tsx
ohltyler, line 1: Please see original comments on this file:
https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/450/files#r1161933922
https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/450/files#r1161933176

public/utils/contextMenu/getActions.tsx
amitgalitz: I can add this and the documentation link part in my PR btw too. same goes for address this comment: https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/482/files#r1195483706

public/components/FeatureAnywhereContextMenu/AnywhereParentFlyout/AnywhereParentFlyout.tsx
amitgalitz: Jackie can probably add more info or give it a try but I was playing around with typesafing this file and it wasn't too straightforward and I had to do something like what I pasted below (not completely done).

If you want to get it working @jackiehanyang feel free to do so or I can add typesafety in my associated detectors PR since I already have context and it might make more sense as I add an additional flyout type. Also let me know if I am over complicating it or doing it wrong with my intution below @ohltyler but the reason for the difficulty is because we instantiate the AnywhereParentFlyout with different files in the final product like the associated flyout and the existing flyout and create.
```
const Flyout: ({ embeddable, closeFlyout, setMode }: AssociatedDetectorsProps | ExistingDetectorsProps) => JSX.Element = {
    associated: AssociatedDetectors,
    existing: CreateAnomalyDetector,
  }[mode];
```

and also instantiate container like this
```
export const Container = ({ startingFlyout, ...props }: { startingFlyout: keyof FlyoutType, [key: string]: any }) => {
```

ohltyler: Got it. Even just having a `AnywhereParentFlyoutProps` with some optional fields (depending on if instantiated with association or existing) I think would be fine here. But having some idea of what `props` is should be visible so it's much easier to read.

public/utils/contextMenu/getActions.tsx
jackiehanyang: Added props here, tried to define the return type here but I find it really hard to do it. Because the getActions is pushing down an array to createAdActions, and createADActions push it down to createAction in the core dashboard which is a generic functions. It seems like this function is designed to work with multiple types and adapt their behavior accordingly.

public/utils/contextMenu/getActions.tsx
```diff
@@ -36,7 +41,6 @@ export const getActions = ({ core, plugins }) => {
                 {...{
```
ohltyler, line 41: This spread operator can be removed. In other words, just have `{ <content> }` instead of `{ ... { <content> } }`

public/utils/contextMenu/getActions.tsx
```diff
@@ -1,26 +1,31 @@
...
     getDisplayName: () => 'Anomaly Detector',
-    getIconType: () => 'apmTrace',
+    getIconType: () => APM_TRACE,
   },
 ];
 
-export const getActions = ({ core, plugins }) => {
+interface GetActionsProps {
+  core: CoreSetup;
+}
```
ohltyler, line 26: `core` should also be consumed as a service similar to `embeddable`, so it doesn't need to be passed as props. If it's some specific service within `core`, that is even better. (e.g., if only notifications service from core is needed, can do something like `setNotifications(core.notifications)`, and then access via `getNotifications()`)

public/components/FeatureAnywhereContextMenu/AnywhereParentFlyout/AnywhereParentFlyout.tsx
ohltyler, line 10: Would still prefer to have some indication of what `props` is here. See comment here: https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/482#discussion_r1195612865

public/utils/contextMenu/getActions.tsx
ohltyler: Got it - i see similar imports for this component in OSD, sounds good.

public/utils/contextMenu/getActions.tsx
jackiehanyang: tried to do something like this `const startServices = createStartServicesGetter(core.getStartServices);` in the plugin.ts file to avoid passing core. However, I don't find a getter function for it - https://github.com/search?q=org%3Aopensearch-project%20createStartServicesSetter&type=code

ohltyler: What service do you need from core? It should be exposed and you can just set that service of core.

jackiehanyang: removed passing core as a prop, added Overlays setter and getter

public/plugin.ts
```diff
@@ -61,7 +63,7 @@ export class AnomalyDetectionOpenSearchDashboardsPlugin implements
       setClient(core.http);
 
       // Create context menu actions. Pass core, to access service for flyouts.
```
ohltyler, line 65: nit: can update this comment

ohltyler (approved):
LGTM! just one nit comment

jackiehanyang:
> LGTM! just one nit comment

will address this in the next pr

kaituo (approved):

PR #487, opened 2023-05-20
Feature anywhere create detector flyout page

jackiehanyang:
### Description

- the pr introduces the create detector flyout page.

#### in the new revision:
- moved some functions to a helper function file
- updated the saved object part to make it suite the latest saved object commits in OSD. Also including extract OVERLAY_ANOMALIES to a constant, referring PointInTimeEvents is a VisLayerType from vis_arguments plugin

#### Missing stuff that coming up in the next pr:
- UTs
- Setup Alert button in the toast
- Update current success toast message after the Setup Alert button is up
- Update plugin.ts file to move dependency plugins from start to setup

### Issues Resolved

### Check List

  - [ ] All tests pass


public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,823 @@
...
+import {
+  AD_DOCS_LINK,
+  MAX_FEATURE_NUM,
+} from '../../../../public/utils/constants';
+
+function AddAnomalyDetector({
+  embeddable,
+  closeFlyout,
+  notification,
+  services,
```
ohltyler, line 95: We should use `getNotifications()` from `services` file instead of passing down`notification` as a prop.

And i think `services` can be removed, I don't see it used in the file

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,823 @@
...
+
+  const [isShowVis, setIsShowVis] = useState(false);
+  const [accordionsOpen, setAccordionsOpen] = useState({ modelFeatures: true });
+  const [detectorNameFromVis, setDetectorNameFromVis] = useState(
+    formikToDetectorName(embeddable.vis.title)
+  );
+  const [intervalValue, setIntervalalue] = useState(10);
+  const [delayValue, setDelayValue] = useState(1);
+  const [enabled, setEnabled] = useState<boolean>(false);
+  const [shingleSizeValue, setShingleSizeValue] = useState(8);
```
ohltyler, line 120: These values should be coming from formik fields instead of using react state. Then you can remove all of `on*Change()` fns below. You already are doing it for the features so you should be able to extend it to the rest of the values as well.

Basically the component can be wrapped in a `Field` with the formik field value and populated automatically. See https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/public/pages/ConfigureModel/components/AdvancedSettings/AdvancedSettings.tsx#L61 as an example

ohltyler: Additionally all of these fields can be added as part of the validation, so any field errors etc is highlighted automatically when submitting. Example: https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/public/pages/ConfigureModel/containers/ConfigureModel.tsx#L144

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,823 @@
...
+  const onDelayChange = (e, field) => {
+    field.onChange(e);
+    setDelayValue(e.target.value);
+  };
+  const aggList = embeddable.vis.data.aggs.aggs.filter(
+    (feature) => feature.schema == 'metric'
+  );
+  const featureList = aggList.filter(
+    (feature, index) => index < (aggList.length < 5 ? aggList.length : 5)
+  );
```
ohltyler, line 145: Was there any UX decision on displaying any messaging if metrics limit > feature limit? 

Also, we should actually be pulling the feature limit from cluster settings instead of hardcoded to 5, but that will be an overall improvement since we have it hardcoded elsewhere too so no blocker on that part.

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,823 @@
...
+
+        const savedObject = await createAugmentVisSavedObject(
+          savedObjectToCreate
+        );
+
+        const saveObjectResponse = await savedObject.save({});
+      });
+      closeFlyout();
+    } catch (e) {
+    } finally {
```
ohltyler, line 202: We should propagate some error response here. There could be errors during the creation, during the start, or during the saving of the saved object. We have existing logic on the AD create page: https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/public/pages/ReviewAndCreate/containers/ReviewAndCreate.tsx#L235

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,823 @@
...
+      formikHelpers.setSubmitting(false);
+    }
+  };
+
+  function formikValueToDetector(values): Detector {
```
ohltyler, line 207: You may use the existing helper fn here: https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/public/pages/ReviewAndCreate/utils/helpers.ts#L29

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,823 @@
...
+    shingleSize: shingleSizeValue,
+    filterQuery: { match_all: {} },
+    description: '',
+    resultIndex: undefined,
+    filters: [],
+    featureList: featureListToFormik(featureList),
+    categoryFieldEnabled: false,
+    realTime: true,
+    historical: false,
+  };
```
ohltyler, line 295: Some of these values we can use the existing default values for. For example, initial formik values on the existing creation flow are here:
detector definition values: [link](https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/public/pages/DefineDetector/utils/constants.tsx#L36)
model configuration values: [link](https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/public/pages/ConfigureModel/utils/constants.tsx#L19)

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,823 @@
...
+
+  const handleAssociate = async (detectorId: string) => {
+    console.log('inside handleAssociate');
```
ohltyler, line 298: can remove

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,823 @@
...
+        detectorId: detectorId,
+      },
+    } as VisLayerExpressionFn;
+
+    const savedObjectToCreate = {
+      title: 'test-title',
+      pluginResourceId: detectorId,
+      visId: embeddable.vis.id,
+      visLayerExpressionFn: fn,
+    } as ISavedAugmentVis;
```
ohltyler, line 316: The latest `ISavedAugmentVis` interface has changed. See [here](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/4006/files#diff-f3a7084285f53c21fad7d46ef7e11a91926e5568f3ac0c9ef9df96f015f99f2a)

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,823 @@
...
+    categoryFieldEnabled: false,
+    realTime: true,
+    historical: false,
+  };
+
+  const handleAssociate = async (detectorId: string) => {
+    console.log('inside handleAssociate');
+    enum VisLayerTypes {
+      PointInTimeEvents = 'PointInTimeEvents',
+    }
```
ohltyler, line 301: This should be imported from visAugmenter plugin where it's defined.

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,823 @@
...
+    const savedObjectToCreate = {
+      title: 'test-title',
+      pluginResourceId: detectorId,
+      visId: embeddable.vis.id,
+      visLayerExpressionFn: fn,
+    } as ISavedAugmentVis;
+
+    const savedObject = await createAugmentVisSavedObject(savedObjectToCreate);
+
+    const response = await savedObject.save({});
```
ohltyler, line 320: We should do error handling here as well if this saving fails

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,823 @@
...
+        featureId: feature.id,
+        featureName: getFeatureNameFromParams(feature.id),
+        featureEnabled: true,
+        featureType: FEATURE_TYPE.SIMPLE,
+        aggregationBy: 'avg',
+        aggregationOf: [{ label: feature.params.field.name }],
+        aggregationQuery: formikToAggregation(feature),
+      };
+    });
+  }
```
ohltyler, line 371: These should all exist in some form already in helper files. For example, https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/public/pages/ReviewAndCreate/utils/helpers.ts

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,823 @@
...
+
+  return (
+    <div className="add-anomaly-detector">
+      <Formik
+        initialValues={initialDetectorValue}
+        onSubmit={handleSubmit}
+        validateOnChange={false}
```
ohltyler, line 378: Why have this be false?

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,823 @@
...
+                <EuiSpacer size="m" />
+                {mode === 'create' && (
+                  <div className="create-new">
+                    <EuiText size="xs">
+                      <p>
+                        Create and configure an anomaly detector to
+                        automatically detect anomalies in your data and to view
+                        real-time results on the visualization.{' '}
+                        <a
+                          href="https://opensearch.org/docs/latest/observing-your-data/ad/index/"
```
ohltyler, line 435: Can use existing constants for this https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/public/utils/constants.ts#L54

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,823 @@
...
+                      />
+                    </div>
+                    <EuiSpacer size="l" />
+                    <EuiTitle size="s">
+                      <h3>Detector details</h3>
+                    </EuiTitle>
+                    <EuiSpacer size="m" />
+                    {/* {!index && <EuiLoadingSpinner size="l" />} */}
+                    {/* Do not initialize until index is available */}
+
```
ohltyler, line 481: can remove

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,823 @@
...
+                        subTitle="Split a single time series into multiple time series based on categorical fields."
+                        initialIsOpen={false}
+                        isUsingDivider={true}
+                      >
+                        <EuiText size="s">
+                          <p>
+                            The dashboard does not support high-cardinality
+                            detectors.
+                            <a
+                              href="https://opensearch.org/docs/latest/observing-your-data/ad/index/"
```
ohltyler, line 706: Can use `AD_DOCS_LINK`

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,823 @@
...
+                              Learn more <EuiIcon type="popout" />
+                            </a>
+                          </p>
+                        </EuiText>
+                      </MinimalAccordion>
+                    </EnhancedAccordion>
+
+                    <EuiSpacer size="l" />
+                    <EuiTitle size="s">
+                      <h3>Model Features </h3>
```
ohltyler, line 718: nit - extra space

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
ohltyler: Actually, I see you have `MAX_FEATURE_NUM` constant at least, should use that

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,823 @@
...
+                </EuiFlexItem>
+                <EuiFlexItem grow={false}>
+                  {mode === 'existing' ? (
+                    <EuiButton
+                      fill={true}
+                      data-test-subj="adAnywhereCreateDetectorButton"
+                      isLoading={formikProps.isSubmitting}
+                      onClick={() => handleAssociate(selectedDetectorId)}
+                    >
+                      associate detector
```
ohltyler, line 798: Should be capitalized

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,823 @@
...
+                    </EuiButton>
+                  ) : (
+                    <EuiButton
+                      fill={true}
+                      data-test-subj="adAnywhereCreateDetectorButton"
+                      isLoading={formikProps.isSubmitting}
+                      //onClick={formikProps.handleSubmit}
```
ohltyler, line 805: can remove

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,823 @@
...
+                      onClick={() => {
+                        handleValidationAndSubmit(formikProps);
+                      }}
+                    >
+                      create detector
```
ohltyler, line 810: Should be capitalized

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,823 @@
...
+                    >
+                      <EuiSpacer size="s" />
+                      <EmbeddablePanel
+                        embeddable={embeddable}
+                        getActions={() => Promise.resolve([])}
+                        inspector={{ isAvailable: () => false }}
+                        hideHeader
+                        isRetained
+                        isBorderless
+                      />
```
ohltyler, line 472: You can use the embeddable plugin to use `PanelComponent` instead of `EmbeddablePanel` which requires a bunch more props. For example, see [here](https://github.com/ohltyler/OpenSearch-Dashboards/blob/view-events-2/src/plugins/vis_augmenter/public/view_events_flyout/components/base_vis_item.tsx)

public/utils/contextMenu/getActions.tsx
```diff
@@ -31,10 +35,11 @@ export const getActions = () => {
         toMountPoint(
           <Provider store={store}>
             <AnywhereParentFlyout
-                startingFlyout={startingFlyout}
-                embeddable={embeddable}
-                closeFlyout={() => overlay.close()}
-              />
+              startingFlyout={startingFlyout}
+              embeddable={embeddable}
+              notification={getNotifications()}
```
ohltyler, line 40: You can remove this as a prop and just access notifications when you need it in the child components via `getNotifications()`

ohltyler: Essentially we should never be passing services down in props - they can all just be accessed whenever we need them directly

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
jackiehanyang: There wasn't any UX decision on displaying any messaging, I will double check Xenia on this.

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
jackiehanyang: The feature list to UI metadata conversion is different than the existing helper function, cause when we first built the feature list, we built it based on the embeddable vis param series field, the format is slightly different than the existing feature lit.

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
jackiehanyang: please ignore associate related logics, I will comment them out in the next revision

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
jackiehanyang: I checked the existing helper file while doing this part and tried to reuse the existing functions as much as possible, however, since we do prefill from the vis, so we need some function to convert vis information to feature list. Currently in the existing create flow, this part is handled by the create detector logic - we parse the formik inputs to feature list. So when we display it later, it's in the correct feature list format. But for feature anywhere, we need some new helper function to convert vis feature list information to correct feature list format.

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
ohltyler: At the least, we could modify the helper fn to handle this case so we don't duplicate.

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
jackiehanyang: updated to true

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,823 @@
...
+
+                    <EnhancedAccordion
+                      id="modelFeatures"
+                      title="Features"
+                      isOpen={accordionsOpen.modelFeatures}
+                      onToggle={() => onAccordionToggle('modelFeatures')}
+                    >
+                      <EuiSpacer size="s" />
+
+                      <FieldArray name="featureList">
```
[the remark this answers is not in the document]
jackiehanyang: I'm reusing <FeatureAccordion> part, the reason for not reusing the whole <Feature> component is that we are using the new customized accordion - `EnhancedAccordion` as the displaying style of the Feature component part

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
ohltyler: I'm still a little confused. If you structure the formik fields to be formatted the same as the existing creation formik fields, then that should eliminate needing any formik conversion functions, or needing new components (can reuse old components if the same formik fields are passed to it.

Understood there may be some edge cases to convert some vis values _to_ formik, but that should just be a one time operation, and then everything downstream can use existing logic/components. For this part, I think it would be best to have a single `visToFormik()` fn or something to populate all fields that need to be specially set based on the vis config.

public/utils/contextMenu/getActions.tsx
jackiehanyang: removed `notification={getNotifications()}` and passing it as a prop. Will do `getNotifications()` when we need to use notification service

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
jackiehanyang: removed

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
jackiehanyang: changed to Associate Detector

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
jackiehanyang: removed

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
jackiehanyang: updated with using link constant

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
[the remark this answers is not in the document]
ohltyler: If the only difference is the accordion, I highly suggest to refactor everything within the accordion that already exists as a reusable component. For example, everything within `FieldArray` to me looks to be identical. We could extract that out and then call it as needed in both the existing `Features` component and here. This would save us >50 lines duplicate code that we don't need to maintain twice

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
jackiehanyang: a bit confused on this - if I remove the `useState` usage, then how do I set an initial value/state to the field? The example you mentioned will render an empty field instead of showing an initial value when load the page

ohltyler: You can set defaults in `initialDetectorValue`. This is what you already pass anyways as the initial values in the `Formik` component (line 376).

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
jackiehanyang: updated it with more error responses during the creation and the start. Not sure about the error wording for the saved object, will leave it to Amit @amitgalitz

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/styles.scss
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
```
amitgalitz, line 4: In order to display the embeddable correctly we should add styling to the vis similar to this:https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/commit/317b33676af31e67df2120804b85d8a146fa5be9#diff-a9719e85099dbb53d9d031c8ba22bdfcae1606179e6debf74e39b6758311edc3R1

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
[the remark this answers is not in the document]
jackiehanyang: We cannot reuse the whole Feature component just by refactoring the accordion because the existing Feature page is not in accordion format, only the feature list is. So I'm customizing and reusing the feature list part - `<FeatureAccordion>`. For feature anywhere, feature list is an accordion list nested in parent Feature accordion. Also, not everything within the accordion is identical, we're customizing the style of the delete and add feature buttons here as the mockup doesn't look exactly the same as the existing page

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
amitgalitz: I was referred by Ashish to use the `EmbeddableRenderer` instead here like shown [here](https://github.com/opensearch-project/alerting-dashboards-plugin/pull/553). Do you have a preference for that vs `PanelComponent`

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/styles.scss
jackiehanyang: updated in the new revision

public/components/FeatureAnywhereContextMenu/EnhancedAccordion/EnhancedAccordion.tsx
```diff
@@ -0,0 +1,87 @@
...
+  EuiTitle,
+  EuiSpacer,
+  EuiButtonIcon,
+  EuiButtonEmpty,
+  EuiAccordion,
+  EuiPanel,
+} from '@elastic/eui';
+import './styles.scss';
+
+const EnhancedAccordion = ({
```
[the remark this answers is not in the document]
jackiehanyang: haven't seen one yet

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,740 @@
...
+            type: VisLayerTypes.PointInTimeEvents,
+            name: 'overlay_anomalies',
+            args: {
+              detectorId: response.response.id,
+            },
+          } as VisLayerExpressionFn;
+
+          const savedObjectToCreate = {
+            title: embeddable.vis.title,
+            pluginResourceId: response.response.id,
```
amitgalitz, line 193: Make sure to update the fields here like [this](https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/487#discussion_r1202853023).

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,740 @@
...
+    }
+  };
+
+  const handleSubmit = (formikProps) => {
+    formikProps.setSubmitting(true);
+    try {
+      const detectorToCreate = formikToDetector(formikProps.values);
+      dispatch(createDetector(detectorToCreate))
+        .then(async (response) => {
+          notifications.toasts.addSuccess(
```
amitgalitz, line 161: We have different notification messages in the mock ups, we should probably align with what is there and/or bring up with UX if we want to make any changes

amitgalitz:
Can you call out in description (potentially also TODO comments) if there are any misses or things that we should remember in the PR. like if we decide that i should add more error description to saved object or if we need changes in wording that we are waiting on from UX. Also add a screenshot of the create view to the description

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
jackiehanyang: Updated it with EmbeddableRenderer in the new revision

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
amitgalitz: Will do as part of the alerting integration

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
ohltyler: I still see shingle size state, lmk if this is part of a new commit.

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
ohltyler: It should be consistent on both create & associate flows. Maybe you can work together to have a shared function (or at least copied logic) for both of these so they are consistent @jackiehanyang @amitgalitz

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
ohltyler: Got it - that makes sense.

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
ohltyler: Sure, no hard opinion on that, makes sense to me. May be more difficult to try to pull those defaults in, given we need to set so many here already based on the vis. I'm good with that

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,740 @@
...
+                  getErrorMessage(
+                    err,
+                    'There was a problem starting the real-time detector'
+                  )
+                )
+              );
+            });
+          enum VisLayerTypes {
+            PointInTimeEvents = 'PointInTimeEvents',
+          }
```
ohltyler, line 182: See https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/487#discussion_r1200789463

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,740 @@
...
+            visId: embeddable.vis.id,
+            savedObjectType: 'visualization',
+            visLayerExpressionFn: fn,
+          } as ISavedAugmentVis;
+
+          const savedObject = await createAugmentVisSavedObject(
+            savedObjectToCreate
+          );
+
+          const saveObjectResponse = await savedObject.save({});
```
ohltyler, line 203: We still will need to have error handling of this part. Refer to lines 287-311 of [this file](public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx). 

@amitgalitz @jackiehanyang can you work together to have this be consistent across both create and associate flows?

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,740 @@
...
+                  )
+                )
+              );
+            });
+          enum VisLayerTypes {
+            PointInTimeEvents = 'PointInTimeEvents',
+          }
+          const fn = {
+            type: VisLayerTypes.PointInTimeEvents,
+            name: 'overlay_anomalies',
```
ohltyler, line 185: See https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/487#discussion_r1200786597

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,740 @@
...
+        featureEnabled: true,
+        featureType: FEATURE_TYPE.SIMPLE,
+        importance: 1,
+        newFeature: false,
+        aggregationBy: 'sum',
+        aggregationOf: visAggregationToFormik(feature),
+        aggregationQuery: JSON.stringify(visAggregationQueryToFormik(feature)),
+      };
+    });
+  }
```
ohltyler, line 290: Recommend to move these to a helpers.ts file or something similar to (1) declutter this file a bit and (2) be consistent with the rest of the plugin. Otherwise I am ok with these particular conversion functions, they all make sense to me why we can't use existing ones for these particular fields

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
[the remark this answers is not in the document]
ohltyler: Got it, thanks for explaining that a bit more. Ideally we should have pushed UX to keep things consistent with existing to prevent (1) code duplication and (2) UI consistency. But given the current mocks this needs to happen this way I guess.

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
ohltyler: Oh cool - wasn't aware of that. That's a better change

ohltyler: Thanks for updating!

ohltyler:
Additionally, we need to have some unit tests here. I'm very concerned about pushing this huge of changes with no tests, otherwise we don't have confidence or control in how the system will behave.

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
amitgalitz: Yes do you think it should be a part of this PR or I can add this in a separate PR when I add the alerting integration on the notification. For here Jackie could copy what I have done in the associate existing, the link you sent is broken for me but I assume its pointing to my PR in the file AddAnomalyDetector. Jackie could add that and then when I rebase with her changes I can refactor it to a helper fn and make sure they are the same. I can then also add more logic with the set up alerting notification

ohltyler: That sounds like a plan to me :), yes you're correct in the file I was referring to. For some reason those github links have been broken as of late...

public/plugin.ts
```diff
@@ -46,6 +48,7 @@ export interface AnomalyDetectionSetupDeps {
 
 export interface AnomalyDetectionStartDeps {
```
amitgalitz, line 49: I think we might have to move all these to the setup deps because they are being used from the flyout which is rendered on top of OSD. Since the ad-dashboard plugin hasn't been started these don't get called if they are in the start. According change needs to be made in line 92 `start` method

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
jackiehanyang: moved helper functions to a helper file

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
jackiehanyang: updated

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
jackiehanyang: reading it from vis_arguments plugin in the new revision

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,823 @@
...
+          type: VisLayerTypes.PointInTimeEvents,
+          name: 'overlay_anomalies',
+          args: {
+            detectorId: response.response.id,
+          },
+        } as VisLayerExpressionFn;
+
+        const savedObjectToCreate = {
+          title: embeddable.vis.title,
+          pluginResourceId: response.response.id,
```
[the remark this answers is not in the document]
jackiehanyang: updated in the new revision

jackiehanyang:
> Can you call out in description (potentially also TODO comments) if there are any misses or things that we should remember in the PR. like if we decide that i should add more error description to saved object or if we need changes in wording that we are waiting on from UX. Also add a screenshot of the create view to the description

Updated the Pr description with more info

jackiehanyang:
> Additionally, we need to have some unit tests here. I'm very concerned about pushing this huge of changes with no tests, otherwise we don't have confidence or control in how the system will behave.

Will add UTs in the next upcoming pr as it's quick a big page and could take me a while to add them. Wanted to get this merged first to unblock Amit since he needs to add Associated related logic on top of this pr

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
```diff
@@ -0,0 +1,740 @@
...
+                      subTitle={
+                        <EuiText size="m">
+                          <p>
+                            Detector interval: {intervalValue} minutes; Window
+                            delay: {delayValue} minutes
+                          </p>
+                        </EuiText>
+                      }
+                    >
+                      <Field name="name" validate={validateDetectorName}>
```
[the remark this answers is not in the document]
amitgalitz: In our code we call more logic after that method is done, like here: https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/public/pages/DefineDetector/containers/DefineDetector.tsx#L137

public/plugin.ts
amitgalitz: I can do this in a follow up PR to reduce scope of change here

public/expressions/constants.ts
```diff
@@ -9,3 +9,5 @@ export const ORIGIN_PLUGIN_VIS_LAYER = 'anomalyDetectionDashboards';
 export const VIS_LAYER_PLUGIN_TYPE = 'Anomaly Detectors';
 
 export const TYPE_OF_EXPR_VIS_LAYERS = 'vis_layers';
+
+export const OVERLAY_ANOMALIES = 'overlay_anomalies"';
```
ohltyler, line 13: nit: can you consume this constant in `overlay_anomalies.ts` file as well. There is a `name` value there. We can set it to this constant instead of redefining there.

ohltyler:
> > Additionally, we need to have some unit tests here. I'm very concerned about pushing this huge of changes with no tests, otherwise we don't have confidence or control in how the system will behave.
> 
> Will add UTs in the next upcoming pr as it's quick a big page and could take me a while to add them. Wanted to get this merged first to unblock Amit since he needs to add Associated related logic on top of this pr

Ok. As long as we ensure we have them before merging the feature branch to main/2.x. Can consider it a blocker at that time.

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
amitgalitz: Can you either copy the lines I have or add a TODO in code with what I should (better do document so nothing is forgotten)

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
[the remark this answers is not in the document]
amitgalitz: I am okay with either you adding this or I can add it in a follow up PR (feel free to add TODO comment highlighting this in code and resolve if you prefer I just follow up here.)

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/styles.scss
amitgalitz: I don't see this style, am I missing it somewhere?

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
amitgalitz: this is discussed here: https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/487#discussion_r1204688791, resolving this comment

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
amitgalitz: @jackiehanyang I think this was missed

public/components/FeatureAnywhereContextMenu/EnhancedAccordion/EnhancedAccordion.tsx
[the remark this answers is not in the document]
amitgalitz: Can you open an issue in OUI or make sure its tracked somewhere either in alerting or AD?

public/components/FeatureAnywhereContextMenu/CreateAnomalyDetector/AddAnomalyDetector.tsx
jackiehanyang: I have addressed this in the last commit

public/expressions/constants.ts
jackiehanyang: updated `overlay_anomalies.ts` file with this constant

amitgalitz (approved):

public/expressions/overlay_anomalies.ts
```diff
@@ -43,8 +44,6 @@ interface Arguments {
   detectorId: string;
 }
 
-const name = 'overlay_anomalies';
-
 export type OverlayAnomaliesExpressionFunctionDefinition =
   ExpressionFunctionDefinition<'overlay_anomalies', Input, Arguments, Output>;
```
ohltyler, line 48: nit: can you add the constant here too?

amitgalitz (approved):

ohltyler (approved):

PR #503, opened 2023-06-01
Unit tests for expression function and additional components

amitgalitz:
### Description

Adding unit tests to the new custom expression function for overlaying anomalies along with UT for `ConfirmUnlinkDetectorModal` and `EmptyAssociatedDetectorMessage` components.

Also made some additional bug fixes:
1. Previously I returned FETCH_FAILURES instead of RESOURCE_DELETED_FAILURES (added unit tests for this) 
2. We know fetch anomalies from custom result index as well, tested this manually as the unit tests itself just mocks the response so not a true test here.
3. Changed context menu option from "Associated anomaly detector" to "Associated detectors"
4. Added "Visualization: " before <viz-title> in associated detectors.
5. changed the associated detectors icon
6. changed the "Associated anomaly detectors" title size to match mock.


public/components/FeatureAnywhereContextMenu/AssociatedDetectors/components/__tests__/ConfirmUnlinkDetectorModal.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+      'Removing association unlinks test-detector-1 detector from the visualization but does not delete it. The detector association can be restored.'
+    );
+    expect(container).toMatchSnapshot();
+  });
+  test('should call onConfirm() when closing', async () => {
+    const { container, getByText, getByTestId } = render(
+      <ConfirmUnlinkDetectorModal {...ConfirmUnlinkDetectorModalProps} />
+    );
+    getByText('Remove association?');
+    await waitFor(() => {});
```
ohltyler, line 52: What does this do exactly? Does this wait a specified amount of time?

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/components/__tests__/__snapshots__/ConfirmUnlinkDetectorModal.test.tsx.snap
```diff
@@ -0,0 +1,22 @@
...
+  data-aria-hidden="true"
+/>
+`;
+
+exports[`ConfirmUnlinkDetectorModal spec renders the component correctly 1`] = `
+<div
+  aria-hidden="true"
+  data-aria-hidden="true"
+/>
+`;
```
ohltyler, line 22: Modals and snapshots don't get along with jest, so I think we can remove the `toMatchSnapshot()` calls in the tests for these. Snapshots here are empty and don't contain anything useful

public/pages/utils/__tests__/constants.ts
```diff
@@ -134,3 +134,171 @@ export const selectedDetectors = [
...
+    hits: {
+      total: {
+        value: 255,
+        relation: 'eq',
+      },
+      max_score: 0,
+      hits: [
+        {
+          _index: '.opendistro-anomaly-results-history-2022.05.06-1',
+          _id: 'hNX8l4ABuV34PY9I1EAZ',
```
ohltyler, line 158: nit: given you have the ID hardcoded in the tests a bunch, could be useful to pull this into a constant as well to reference within here + in all of the tests that are using the same ID

public/expressions/overlay_anomalies.ts
```diff
@@ -66,7 +66,7 @@ const getAnomalies = async (
     endTime,
     detectorId,
     undefined,
-    false
+    true
```
ohltyler, line 69: this needs to revert to `false` right?

public/expressions/__tests__/overlay_anomalies.test.ts
```diff
@@ -0,0 +1,152 @@
...
+          parsedAnomalies,
+          ADPluginResource
+        );
+      expect(pointInTimeEventsVisLayer).toStrictEqual(
+        expectedPointInTimeEventsVisLayer
+      );
+    });
+  });
+  describe('getErrorLayerVisLayer()', () => {
+    test('get resource deleted ErrorVisLayer', async () => {
```
ohltyler, line 117: nice - really good to have these tests

public/expressions/overlay_anomalies.ts
```diff
@@ -79,13 +79,13 @@ const getAnomalies = async (
   return parsePureAnomalies(anomalySummaryResponse);
 };
 
-const getDetectorResponse = async (detectorId: string) => {
+export const getDetectorResponse = async (detectorId: string) => {
```
ohltyler, line 82: nit: you could pull out all of these helper fns used in the expression fn into a separate file and call them in here, to keep the function file a little cleaner and more concise. This is what we do on OSD side. But no strong feelings either way, i'm fine with the current way too

public/expressions/overlay_anomalies.ts
```diff
@@ -103,6 +103,56 @@ const convertAnomaliesToPointInTimeEventsVisLayer = (
...
+          : error instanceof Error
+          ? get(error, 'message', '')
+          : '',
+    };
+  } else if (
+    typeof error === 'string'
+      ? error.includes(DETECTOR_HAS_BEEN_DELETED)
+      : get(error, 'message', '').includes(DETECTOR_HAS_BEEN_DELETED)
+  ) {
+    console.log('inside resource deleted');
```
ohltyler, line 132: can remove

public/expressions/overlay_anomalies.ts
```diff
@@ -103,6 +103,56 @@ const convertAnomaliesToPointInTimeEventsVisLayer = (
...
+        error === 'string'
+          ? error
+          : error instanceof Error
+          ? get(error, 'message', '')
+          : '',
+    };
+  } else if (
+    typeof error === 'string'
+      ? error.includes(DETECTOR_HAS_BEEN_DELETED)
+      : get(error, 'message', '').includes(DETECTOR_HAS_BEEN_DELETED)
```
ohltyler, line 130: nit: you could pull this out into a helper fn like you do for `checkIfPermissionErrors()` - something like `checkIfDeletionErrors()`

public/pages/utils/__tests__/constants.ts
```diff
@@ -134,3 +134,171 @@ export const selectedDetectors = [
     lastUpdateTime: 1651818220194,
   },
 ] as DetectorListItem[];
+
+export const anomalyResultSummary = {
```
ohltyler, line 138: nit: I prefer all constant vars to be in ALL_CAPS so its easy to differentiate and makes the tests where these constants are used a bit more readable

public/components/FeatureAnywhereContextMenu/AssociatedDetectors/components/__tests__/ConfirmUnlinkDetectorModal.test.tsx
amitgalitz: To be honest this shouldn't really add any time because the async function within is empty but this has been helpful in other tests where we execute some action and we want some wait as I assume just the waitFor function on its own adds some fractional wait time. However in this case since I am only really looking for the `toHaveBeenCalled()` to be true and not a different rendering I don't think its needed at all. The tests pass without it too.

public/pages/utils/__tests__/constants.ts
amitgalitz: made the change

public/expressions/overlay_anomalies.ts
amitgalitz: changed it back, good catch

ohltyler (approved):
LGTM!

amitgalitz:
UT are passing all locally but failing here cause its running against main. Unless any opposition I change it to run against feature-anywhere branch for this branch.

public/expressions/helpers.ts
```diff
@@ -31,22 +31,32 @@ import { get } from 'lodash';
...
     undefined,
-    false
-  );
-
-  const anomalySummaryResponse = await getClient().post(
-    `..${AD_NODE_API.DETECTOR}/results/_search`,
-    {
-      body: JSON.stringify(anomalySummaryQuery),
-    }
+    true
```
ohltyler, line 42: Don't forget to revert this again back to `false`

amitgalitz: done

ohltyler (approved):

jackiehanyang (dismissed):

ohltyler (approved):

jackiehanyang (approved):

PR #585, opened 2023-08-07
Fix display of detector names

smortex:
When starting or stopping detectors, the pop-up window shows the detector names as `DETECTOR_NAME />` where "DETECTOR\_NAME" is the name of the detector.

The markup seems to have residual `/>` that are rendered to the user. Remove them to ensure they do not appear between the opening and the closing EuiLink tags.

screenshot of the issue

To reproduce:
1. Visit https://playground.opensearch.org/app/anomaly-detection-dashboards#/detectors?from=0&size=20&search=&indices=&sortField=name&sortDirection=asc
2. Check a few detectors
3. Click on Actions > Stop real-time detectors


kaituo (approved):

smortex:
CI failures seems unrelated :roll_eyes: :

* https://github.com/opensearch-project/anomaly-detection does not build;
* `yarn install` failed on a network error (transient failure, may be fixed by re-starting the test).

jackiehanyang (approved):

PR #616, opened 2023-10-09
Prevent empty task IDs passed to server side

ohltyler:
### Description

This PR improves the intermediate loading state on the historical results page. Before, an initial load of the page would send requests to the server-side with an empty task ID, which would return a 404 not found due to the path not existing, and show empty results. This is because there would be a double slash `//` where the task ID was supposed to be in between- e.g., `/api/anomaly_detectors/detectors//results/` (_note this would only happen if the page was directed via direct URL or refresh, or from the 'View results' button on the detector list page. If navigating from the detector results tab to the historical tab, the issue was not seen since the detector details were already populated in redux store_). Typically, this was quickly re-executed with a populated task ID after it was asynchronously fetched, and re-render the page with populated results. Because of this, it was not found as an issue, and was only noticeable when analyzing the network requests from a browser dev tools extension.

To simply prevent these requests from happening at all, we wrap all network requests with an `isValidToFetch()` fn to prevent empty task IDs from being propagated.

#### Testing
I've confirmed that results load as normally for HC and non-HC, real-time and historical. Now, the 404 errors are never seen from dev tools since the invalid requests are blocked from occurring in the first place.

### Check List

### DIFF
--- a/public/pages/DetectorResults/containers/AnomalyHistory.tsx
+++ b/public/pages/DetectorResults/containers/AnomalyHistory.tsx
@@ -173,6 +173,12 @@ export const AnomalyHistory = (props: AnomalyHistoryProps) => {
   const backgroundColor = darkModeEnabled() ? '#29017' : '#F7F7F7';
   const resultIndex = get(props, 'detector.resultIndex', '');
 
+  // Utility fn to only fetch data when either it is non-historical, or historical and
+  // there is a populated task ID which will be used to fetch the historical results
+  const isValidToFetch = () => {
+    return !props.isHistorical || (props.isHistorical && taskId.current);
+  };
+
   // Tracking which parent category fields the user has selected to filter by.
   const [selectedCategoryFields, setSelectedCategoryFields] = useState(
     getCategoryFieldOptions(detectorCategoryField)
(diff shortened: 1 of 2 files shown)


joshpalis (approved):

public/pages/DetectorResults/containers/AnomalyHistory.tsx
```diff
@@ -173,6 +173,12 @@ export const AnomalyHistory = (props: AnomalyHistoryProps) => {
   const backgroundColor = darkModeEnabled() ? '#29017' : '#F7F7F7';
   const resultIndex = get(props, 'detector.resultIndex', '');
 
+  // Utility fn to only fetch data when either it is non-historical, or historical and
+  // there is a populated task ID which will be used to fetch the historical results
+  const isValidToFetch = () => {
```
amitgalitz, line 178: if it is not hisotorical even on first load there will be a detector ID?

ohltyler: Yes- the `AnomalyHistory` component is conditionally rendered based on a valid detector, so there will always be a detector ID. (additionally it is required in the url path which cannot be empty). The edge case is that the task ID is asynchronously fetched after the detector ID (via get detector details API) to populate it, which is why the initial render is happening without a task ID.

A different option is to refactor some of the `HistoricalDetectorResults` page to try to prevent rendering the entire `AnomalyHistory` component until the task ID is valid. But the current way may provide a better user experience in that it is more react-like, where the only re-rendering is happening in the charts component where the task ID is used to actually populate it once it is fetched and data is retrieved.

amitgalitz (approved):

kaituo (approved):

PR #702, opened 2024-03-04
Remove legacy dependency

ohltyler:
### Description

Remove a legacy dependency not needed by plugins anymore. The consequence is it removes a huge amount of transitive dependencies in `yarn.lock`, many of which are dated and frequently cause CVE-related issues.

Testing done
- `yarn build` works
- `yarn osd bootstrap` works
- sanity test on a local domain

~~Still understanding other potential impacts and original reasoning for having this dependency, will update here when more is learned from @opensearch-project/opensearch-core team.~~ UPDATE: this has been confirmed by @kavilla , no other concerns.

### Check List

  - [x] All tests pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,6 @@
     ]
   },
   "devDependencies": {
-    "@elastic/eslint-import-resolver-kibana": "link:../../packages/osd-eslint-import-resolver-opensearch-dashboards",
     "@testing-library/user-event": "^12.1.6",
     "@types/react-plotly.js": "^2.6.0",
     "@types/redux-mock-store": "^1.0.1",
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -1,7 +1,0 @@
----
-extends: ['@elastic/eslint-config-kibana', 'plugin:@elastic/eui/recommended']
-
-settings:
-  import/resolver:
-    '@elastic/eslint-import-resolver-kibana':
-      rootPackageName: 'anomaly-detection-dashboards'
--- a/yarn.lock
+++ b/yarn.lock
@@ -9,10 +9,6 @@
   dependencies:
     regenerator-runtime "^0.13.11"
 
-"@elastic/eslint-import-resolver-kibana@link:../../packages/osd-eslint-import-resolver-opensearch-dashboards":
-  version "0.0.0"
-  uid ""
-
 "@hapi/hoek@^9.0.0":
   version "9.3.0"
   resolved "https://registry.yarnpkg.com/@hapi/hoek/-/hoek-9.3.0.tgz#8368869dcb735be2e7f5cb7647de78e167a251fb"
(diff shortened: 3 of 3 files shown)


ohltyler:
@opensearch-project/opensearch-core could I get assistance in understanding why this dependency is not needed anymore? My understanding is it was needed somewhere in the build or bootstrap logic for combining dependencies from core + plugin to build/run the plugin successfully. Is there any reference PR in core that I could reference here?

dbwiddis (dismissed):
Nice.

Can we just completely delete `yarn.lock` and add it to `.gitignore` since it's always re-generated at run time anyway?  Or would that prevent the CVE scanners from figuring out what needs updating?

ohltyler:
> Nice.
> 
> Can we just completely delete `yarn.lock` and add it to `.gitignore` since it's always re-generated at run time anyway? Or would that prevent the CVE scanners from figuring out what needs updating?

We still want the lockfile, and with this fix it should not be re-generated often; only when we make direct changes in our package.json. The reason it was much more fluid before is the fact it was pulling in updated dependencies from the tip of main of core OSD.

ohltyler:
@kavilla any further insight on this?

package.json
```diff
@@ -23,7 +23,6 @@
     ]
   },
   "devDependencies": {
-    "@elastic/eslint-import-resolver-kibana": "link:../../packages/osd-eslint-import-resolver-opensearch-dashboards",
```
kavilla, line 26: I would modify: https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/.eslintrc.yml#L6 as well.

From my understanding this dependency has it's use cases but I don't see it being used within this plugin repo that makes it worth linking.

If this was a shared repo in the past, or if you need to extend the imports rules then I get see it. But after removing and modifying the eslintrc, I needed to modify the import orders as I believe our lint rules kicked in for import order and it didn't treat your plugin as the root directory

kavilla (approved):

ohltyler:
Removed lintfile, re-did a bunch of sanity checks, looks good.

UT:
```
Test Suites: 1 skipped, 76 passed, 76 of 77 total
Tests:       2 skipped, 364 passed, 366 total
Snapshots:   70 passed, 70 total
Time:        98.704 s
Ran all test suites.
✨  Done in 100.18s.

```
Build:
```
yarn build
yarn run v1.22.19
$ yarn plugin-helpers build && echo Renaming artifact to $npm_package_config_plugin_zip_name-$npm_package_config_plugin_version.zip && mv ./build/$npm_package_config_plugin_name*.zip ./build/$npm_package_config_plugin_zip_name-$npm_package_config_plugin_version.zip
$ node ../../scripts/plugin_helpers build
 info Loaded config file from [/Users/ohltyler/workspace/opensearch/ad/3.0/OpenSearch-Dashboards/plugins/anomaly-detection-dashboards-plugin-1/.opensearch_dashboards-plugin-helpers.json]
 info deleting the build and target directories
 info running @osd/optimizer
 │ info initialized, 0 bundles cached
 │ info starting worker [1 bundle]
 │ succ 1 bundles compiled successfully after 41 sec
 info copying assets from `public/assets` to build
 info copying server source into the build and converting with babel
 info running yarn to install dependencies
 info compressing plugin into [anomalyDetectionDashboards-3.0.0.zip]
 info cleaning up compression temporary artifacts
Renaming artifact to 
```
Sanity tested functionality and checked all pages on local cluster, no problems.

amitgalitz:
I see some failures in plugin bootstrap in CI, should we address these?

dbwiddis (approved):

ohltyler:
> I see some failures in plugin bootstrap in CI, should we address these?

CI passes for linux. The failure is some timeout issue specific to windows, I believe this is an overall infra issue across the project, needs further investigation. Remote integ test failure is being addressed by @jackiehanyang , Jackie, can you include this in your investigation?

Prefer to keep the scope the same for this PR.

amitgalitz (approved):

Issue #721, opened 2024-04-09
[META] MDS Support For AD


jackiehanyang:
### Implementation 
Pages:
- [x] Dashboard Overview Page `<DashboardOverview>`
- [x] Detector List Page `<DetectorList>` PR - https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/716
- [x] Create Detector Page step1 `<CreateDetectorSteps>` 
- [x] Create Detector Page step2 `<DefineDetector>`
- [x] Create Detector Page step3 `<ConfigureModel>`
- [x] Detector Detail Page `<DetectorDetail>` PR - https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/719
- [x] Anomaly Detection Overview Page `<AnomalyDetectionOverview>`
- [x] Feature Anywhere Page

Redux Store:
- [x] all functions in ad store has been updated with MDS support
- [x] all functions in anomalyResults store has been updated with MDS support
- [x] all functions in alerting store has been updated with MDS support
- [x] all functions in index store has been updated with MDS support
- [x] all functions in liveAnomalyResults store has been updated with MDS support
- [x] all functions in mapper store has been updated with MDS support
- [x] all functions in opensearch store has been updated with MDS support
- [x] all functions in previewAnomalies store has been updated with MDS support
- [x] all functions in sampleData store has been updated with MDS support

Server:

- [x] all api calls in ad route has been updated with MDS support
- [x] all api calls in alerting route has been updated with MDS support
- [x] all api calls in opensearch route has been updated with MDS support
- [x] all api cals in smapleData route has been updated with MDS support

Test:

- [x] update failed existing UTs
- [x] add UTs for server side changes
- [ ] add integration tests

Temp Hardcoded Value
- [x] line 47 in https://github.com/jackiehanyang/anomaly-detection-dashboards-plugin/edit/neoDetailPage/public/pages/CreateDetectorSteps/hooks/useFetchDetectorInfo.ts?pr=%2Fopensearch-project%2Fanomaly-detection-dashboards-plugin%2Fpull%2F719

PR #719, opened 2024-04-07
Support MDS on DetectorDetails page

jackiehanyang:
### Description

This PR adds MDS support to the DetectorDetails page. The page's functionality and appearance remain the same, with the exception of the following changes:
- a dataSourceId will be appended to the url for both `/results`, `/historical`, and `/configurations` tab
- A read-only data source view will be rendered in the top right corner. Currently, it displays a placeholder label as the Neo team is implementing the feature to allow the data source view to take only a dataSourceId instead of requiring both an id and a label.

Please note:
- This PR branch was created from the `neoListPage` branch, which has another open PR. We will aim to merge that PR first to provide a clearer view for this PR.
- Efforts were made to decouple page interactions to allow for incremental PRs. However, hard-coded values were used in some places to ensure functionality. These hard-coded values will be removed in future updates when working on adding MDS support on those specific pages.

Screenshot 2024-04-07 at 08 46 58

### Issues Resolved

[List any issues this PR will resolve]

### Check List

- [ ] New functionality includes testing.
  - [ ] All tests pass
- [ ] New functionality has been documented.
- [ ] Commits are signed per the DCO using `--signoff`

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #722, opened 2024-04-12
Support MDS on List, Detail, Dashboard, Overview pages

jackiehanyang:
### Revisions:
4/17: 
- The latest revision removes dataSourceEnabled check before dispatch calls
- update router path to put all dataSourceId at the end of the path
- handle NEP situation for dataSource plugin and dataSourceManagement plugin getter and setter
- run prettier
- make server client type safe
- ------

4/16: 
The latest revision modifies the way of reading dataSourceId between pages and components. Now it's getting the dataSourceId from the url instead of receiving it as a prop. Also removed dataSourceId from Detector and DetectorListItem interface to avoid setting dataSourceId in loop. Instead, we just append the dataSourceId that was read from the url to the herf link

### Description:

This PR includes:
- adds MDS support on AD Overview page, Dashboard page, List Detectors Page, and Detector Detail page.
- update all route path to make it accept dataSourceId when it has truthy value 
- update side nav bar to make the redirection includes dataSourceId when it has truthy value
- some bug fixes

### Notes:
- Previous prs for List and Detector Detail page: https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/716, https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/719
Addressed all comments in the previous prs but the update routes to allow empty dataSourceId one, included this fix in this pr
- meta issue - https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/issues/721

https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/assets/41348518/8547210e-3698-49c0-ae19-da88049962b1

### Issues Resolved

### Check List

  - [ ] All tests pass


public/anomaly_detection_app.tsx
```diff
@@ -29,13 +34,18 @@ export function renderApp(coreStart: CoreStart, params: AppMountParameters) {
...
   ReactDOM.render(
     <Provider store={store}>
       <Router>
         <Route
           render={(props) => (
             <CoreServicesContext.Provider value={coreStart}>
-              <Main {...props} />
+              <Main
+                dataSourceEnabled={dataSource.dataSourceEnabled}
+                setHeaderActionMenu={params.setHeaderActionMenu}
```
ohltyler, line 46: 1. Can we fetch `dataSourceEnabled` from services? `getDataSourceManagementPlugin().dataSourceEnabled`. Then we can avoid passing extra props
2. Rather than passing down the `setHeaderActionMenu` to all components which adds lots of extra props, can we persist it on initialization? Maybe exploring adding this in redux store somewhere, and can be fetched there. `configureStore` is exposed here to instantiate the client, this may be a good place where that could be added

public/models/interfaces.ts
```diff
@@ -205,6 +205,7 @@ export type Detector = {
   taskState?: DETECTOR_STATE;
   taskProgress?: number;
   taskError?: string;
+  dataSourceId? : string;
```
ohltyler, line 208: Why can't we store the selected data source ID as a redux store item, and only update when a new ID is selected? Then you don't need to pass around ad-hoc `dataSourceId`s. If we use redux middleware for all of the calls made from client-side, the persisted data source ID within that store can handle all of that and no propagating needs to happen at all from client-side.

Thinking long-term with supporting multiple data sources, I could see some places where more fine-grained IDs need to be attached to the data models, so it makes sense to add as an optional field for `Detector` interface. But even then, this could all be persisted in the redux store and those details don't need to be handled from client-side components at all.

ohltyler (commented):
Please see comments. I see an opportunity to reduce a lot of the props propagation happening with persisting things in redux store.

public/anomaly_detection_app.tsx
jackiehanyang: Having props passed down isn't inherently bad. I do agree that too many props being passed down multiple levels might lead to prop drilling, which can make the code harder to understand and maintain. However, in this case, we're not passing lots of extra props, just this one setHeaderActionMenu function as a prop. It's pretty straightforward as it's not being passed down multiple levels, just passed down one level to the route path landing page. So, it doesn't make much difference performance-wise and maintenance-wise I believe. Also, I'm not sure how to add this param to the Redux configureStore, like create a new Redux store for this? Do you have an example of how to do it?

public/models/interfaces.ts
jackiehanyang: Thanks for the suggestion to utilize the redux store for managing the dataSourceId. I understand the benefits of centralizing the state in Redux, especially for consistency and potentially reducing the complexity of propagating the IDs through components.

However, I have some concerns about relying entirely on the Redux store for this purpose. In the future when our application architecture supports multiple active data sources in different components simultaneously, which might lead to complications if we centralize the dataSourceId state. This approach could limit the flexibility needed for some components to operate independently under different contexts, which we anticipate will be a requirement as we scale and introduce more complex features.

Additionally, managing fine-grained control over data sources at the component level allows us to more easily integrate with other parts of our infrastructure that are not as tightly coupled with Redux. It also provides clearer traceability for data flow in complex workflows, which can aid in debugging and maintenance. Our application already have an overloaded redux state and it even crashed the Redux Debug DevTool because the state is too long to parse on some big pages like detector details page.

public/anomaly_detection_app.tsx
ohltyler: 1. Thanks!
2. Got it, since this is already configured in the `AppMountParameters` and will only go one level down, that can be ok.

public/models/interfaces.ts
ohltyler: > This approach could limit the flexibility needed for some components to operate independently under different contexts

Could you explain how this limits it? I'm not saying we never should have component-level context for different subsets of data source IDs, for example, but rather, by persisting the selected data source ID(s) from the top-down menu in some centralized store, we immediately aren't bound/forced to propagate those top-level IDs as props through many components that may not need it at all, but have to pass through to its children.

From my understanding, the current and near-future approach is to start by selecting a single or multiple data source IDs from a top-level picker. My proposal is that by persisting this general top-level data in redux store, we eliminate all pass-through props and data model changes needed to persist those ID(s) on their own.

Especially for the initial release which is bound to at most one data source ID, this completely eliminates the need to pass any props at all. As things could become more complex, we can introduce more IDs & props where appropriate if local, component-level context is needed, for example.

ohltyler: Lmk your thoughts on this too, if you prefer some way based on the current and future plans of MDS and state management @huyaboo @BionIT

jackiehanyang: let's allow the Neo team to decide if centralized state management suits the neo goal and could benefit the future Neo roadmap. This should be an implementation requirement for all plugins, if necessary. 

Regardless, I considered storing this ID in a Redux store and discussed this solution with Lu initially. The reason I didn't proceed is based on my current experience with adding MDS support for AD, and these concers:
- **Increased complexity**: Implementing Redux adds layers of complexity to the application. And redux often requires writing more boilerplate code, for example I have to setup actions, reducers, selectors, and middleware to manage just a simple ID. 
- **Performance overhead**: Redux connects components to the global store, which can lead to unnecessary re-renders if not carefully optimized. This is already happening to AD, almost all pages gets re-rendered multiple times when opening it.
- **Debugging difficulty**: Debugging Redux can be more challenging than simple prop passing. To understand how and why the dataSourceId is changing, we have to rely on Redux DevTools, as opposed to simply looking at component props. However, as I mentioned earlier regarding the performance overhead, Redux DevTools crashes on some AD pages because the state is too long to serialize. This leaves us with no mechanism to effectively debug the issue. 

Therefore, the reason I choose to make dataSourceId a prop include simplicity, less overhead, easier debugging and testing, reduced coupling. I do agree that when it's used by many components at different nesting levels, a Redux store is definitely the better choice. However, I believe that only the detector detail page has a really nested structure. Additionally, the dataSource picker is disabled on the detector detail page, so there's no need to pass dataSourceId around there. For the create detector page, we only allow customers choose it in the first step. It becomes a read-only value in subsequent steps, so there are no nesting concerns on the create page.

ohltyler: > Increased complexity: Implementing Redux adds layers of complexity to the application. And redux often requires writing more boilerplate code, for example I have to setup actions, reducers, selectors, and middleware to manage just a simple ID.

When done correctly, it actually simplifies the complexity, since it reduces a lot of persisted/shared/common state management that otherwise needs to be handled in the individual components and passed down as props. It's not a lot of boilerplate to add this state - for example, we persist just the `requesting` and `errorMessage` vars for creating sample data [here](https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/public/redux/reducers/sampleData.ts). We could do basically the same thing, except it would be even simpler, and we only need to persist the id. We don't even need to make any server-side calls. Then, it's just a one line change to add the reducer as part of the global reducer.

> Performance overhead: Redux connects components to the global store, which can lead to unnecessary re-renders if not carefully optimized. This is already happening to AD, almost all pages gets re-rendered multiple times when opening it.

It is by design that there is multiple re-renders, they are not unnecessary. The elegant part about redux store is while many different calls may be happening on the server side, the client-side components automatically update when these values change, such as loading/requesting state, the total set of detectors, fetched index mappings, etc. Because we have fine-grained state, this leads to many re-renders, yes. Additionally, adding one field to subscribe to the data source ID wouldn't re-render anymore than what would be done via component-managed state.

> Debugging difficulty: Debugging Redux can be more challenging than simple prop passing. To understand how and why the dataSourceId is changing, we have to rely on Redux DevTools, as opposed to simply looking at component props. However, as I mentioned earlier regarding the performance overhead, Redux DevTools crashes on some AD pages because the state is too long to serialize. This leaves us with no mechanism to effectively debug the issue.

Can you explain what type of debugging is needed exactly? Persisting an ID would be very straightforward - at the top-level of pages, set the data source ID from the URL into redux store once. Now, any component can simply read that ID from store at any point - it would only be removed when client-side store is wiped out (refreshing the page). It's even simpler in that there is no server-side calls being made and nothing else needs to be tracked.

I understand how some of the complexity comes about, where the code is not totally consistent in managing some state at component-level, and some at redux level. But adding a single central state for this ID will simplify things a lot and reduce 100% of props passing. It can be equally useful when this plugin expands to support multiple datasources at once, and will lead to a lot less code changes.

I don't see the current approach as scalable, and we should avoid needing to find workaround solutions like needing to append `dataSourceId` to the `DetectorListItem` interface, for example. We can set up the small amount of infrastructure now and it will make things much simpler as MDS evolves and our plugin integrates more with it.

owaiskazi19: +1 to @ohltyler suggestion here to add dataSourceId to redux store. The reason being, it would be simple to pass dataSourceId to various components. And dataSourceId would be the just first param for the store and later on we can have more. If not now, can we create an issue to refactor this later.

ohltyler: I'm hesitant to have an issue and refactor later. Reason being, we don't want to add a field in a large number of places with plans to immediately remove them. This can lead to hard-to-find bugs if some places are missed, NPE issues, and the general difficulty of if those components or data models are updated after this change, but before it is removed. It could also potentially end up in ui metadata fields at some point on cluster-side and lead to BWC parsing issues.

jackiehanyang: If we really dislike the idea of passing `dataSourceId` as a prop, I can change it to be read from the URL. There are two possible workflows (listed below). Regardless of which one we encounter, reading `dataSourceId` from the Redux store is completely unnecessary and over-engineered.

Workflow 1: When `dataSourceId` is not present in the URL:

1. We first render the page with a dataSource picker and get the default `dataSourceId` from the dataSource picker.
2. We update the URL with this `dataSourceId`.
3. (If we want to store this ID in the Redux store) We update the Redux store with this `dataSourceId`.


Workflow 2: When `dataSourceId` is present in the URL (e.g., a customer bookmarked the page):

1. We render the page with the given URL, which means we already have a `dataSourceId`.
2. We render the dataSource picker and make all the API calls with this `dataSourceId` to render the page.

In these two workflows, reading `dataSourceId` from the Redux store is completely needless. It doesn't serve as a source of truth because the URL is the only source of truth. We render the page based on what is present in the URL, so why waste memory storing this ID in the Redux store, which interacts with 5+ reducers?

huyaboo: My $0.02:

From data source side, there are certain reasons to have state management:
1. **We recommend the use of states to reduce network calls for certain components.** For instance, if a `DataSourceView` component is repeatedly being used for different pages, it is recommended to keep both the `id` and `label` values and pass them into the component. This saves the component a network call and potentially a re-render of the page. Now this would require setting the state of both an `id` and a `label` and this might not be relevant to AD since the `DataSourceView` is used only in the detector detail page.
2. **At the plugin-level, how deeply/how many times is `dataSourceId` being passed in?** If it's being referenced frequently, having to check for `dataSourceId` being passed into every component is more tedious than simply fetching from the redux store.
3.  As for a state management requirement, this should be **case-by-case**. Several plugins only have one/a few pages to pass around the `dataSourceId`. It would be unreasonable to require state management for a prop that's being passed around a handful of times.

ohltyler: Something that would be really cool is having some way to persist the IDs and make them accessible for all of the redux fns that make server-side calls that end up making cluster calls. Then all of the redux calls made on client-side wouldn't even need to pass them through; the middleware could pick up the persisted ID and pass it directly

jackiehanyang: > Please avoid saying this; using redux to maintain common/global/shared state is a popular, practical, and clean solution.

I agree redux is a good library, I'm not saying it's useless :) Just the reason I think it's unnecessary is because that the dataSourceId in the URL is our primary source of truth for identifying a current resource or view, using it directly can simplify our state management. 

If we are retrieving the dataSourceId from the URL on every different page, why not just use it directly instead of following a cycle of use -> update redux store -> read from redux store -> use? For every child component on each page, we could either read the dataSourceId from the Redux store that was maintained on the previous parent page, or simply read the dataSourceId directly from the previous parent page. If MDS adds more attributes in the future, these will also be in the URL. Therefore, I believe that no approach is inherently better than another; they are just different ways of achieving the same outcome. 

The reason I didn't choose the Redux store approach when adding MDS support on the Overview page is due to an existing bug that caused an infinite loop in the global Redux store state changes. This took up more of my time than usual for debugging because you need to understand how every global state is changing behind the scenes. 

Regardless of the approach, we need to set the dataSourceId in DetectorListItem due to this being an interface change. The only difference is whether we set it with the value read from the URL or the value obtained from the Redux store.

I can make the change if this is really a blocker. Guess it's time to disagree and commit :)

ohltyler: Maybe there's still confusion. I'm not saying we don't use it directly - I'm saying that by persisting it, we don't need to have any props passed down to any component. See previous comment:
> The whole point of having a redux store is it can be set once (either from url or from the data source picker), and then is available for any component to use, with no extra state or variables needing to be managed or passed around, at all

The idea is it gets set only the few times at the top-level of pages where its parsing from the URL or from the data source picker. Then any component throughout the entire client-side of the plugin can use that URL - no props needed or needing to find workarounds to propagate the ID through child components. The best benefit is it reduces all of that code and makes it much more flexible, should we need to persist more or different data related to MDS in the future without needing to update interfaces or prop data types.

public/pages/main/Main.tsx
```diff
@@ -23,42 +23,58 @@ import { DefineDetector } from '../DefineDetector/containers/DefineDetector';
...
+  const existingParams = "from=0&size=20&search=&indices=&sortField=name&sortDirection=asc";  
+  const constructHrefWithDataSourceId = (basePath, existingParams, dataSourceId) => {
+    const fullUrl = `${window.location.origin}${basePath}?${existingParams}`;
+    const url = new URL(fullUrl);
+    url.searchParams.set('dataSourceId', dataSourceId); 
+    
+    return `#${url.pathname}${url.search}`;
+  };
+  const dashboardHref = dataSourceId ? `#${APP_PATH.DASHBOARD}?dataSourceId=${dataSourceId}` : `#${APP_PATH.DASHBOARD}`;
+  const overviewHref = dataSourceId ? `#${APP_PATH.OVERVIEW}?dataSourceId=${dataSourceId}` : `#${APP_PATH.OVERVIEW}`;
```
ohltyler, line 60: can we remove these entirely and just use the fn you already define on line 52?

ohltyler: Suppose tomorrow the dashboard and overview pages get optional params added, now this will remove those. The fn you define is more flexible.

public/pages/main/Main.tsx
```diff
@@ -122,22 +147,29 @@ export function Main(props: MainProps) {
                 />
                 <Route
                   path={APP_PATH.OVERVIEW}
-                  render={() => (
+                  render={(props: RouteComponentProps) => (
                     <AnomalyDetectionOverview
-                      isLoadingDetectors={isLoadingDetectors}
```
ohltyler, line 127: Have you confirmed the impact of removing this? Do we still have proper loading state when loading the page and creating sample detectors? Adding details in the PR for this may be helpful.

public/pages/utils/helpers.ts
```diff
@@ -112,3 +112,26 @@ export const formatNumber = (data: any) => {
...
+export const getAllDetectorsQueryParamsWithDataSourceId = (
+  dataSourceId: string
+) => ({
+  from: 0,
+  search: '',
+  indices: '',
+  size: MAX_DETECTORS,
+  sortDirection: SORT_DIRECTION.ASC,
+  sortField: 'name',
+  dataSourceId: dataSourceId,
```
ohltyler, line 125: nit: when field names and var names are the same, you don't need to add the extra definition, only need to write it once:
`dataSourceId: dataSourceId` -> `dataSourceId` 

In fact, with linting enabled, it will throw warnings for this.

public/pages/utils/helpers.ts
```diff
@@ -112,3 +112,26 @@ export const formatNumber = (data: any) => {
...
+
+export const getSampleDetectorsQueryParamsWithDataSouceId = (dataSourceId: string) => (
+  {
+  from: 0,
+  search: 'sample',
+  indices: '',
+  size: MAX_DETECTORS,
+  sortDirection: SORT_DIRECTION.ASC,
+  sortField: 'name',
+  dataSourceId: dataSourceId,
```
ohltyler, line 136: same as above comment

public/services.ts
```diff
@@ -49,4 +60,5 @@ export default {
   getOverlays,
   setUISettings,
   setQueryService,
+  getSavedObjectsClient,
```
ohltyler, line 63: nit: can we add the other added getters here? To make mocking easier for tests.

server/utils/helpers.ts
```diff
@@ -81,3 +85,23 @@ export const prettifyErrorMessage = (rawErrorMessage: string) => {
...
   }
 };
+
+export function getClientBasedOnDataSource(
+  context: any,
+  dataSourceEnabled: boolean,
+  request: OpenSearchDashboardsRequest,
+  dataSourceId: string,
+  client: any
+): (
```
ohltyler, line 95: from following the codepath, the client could be `undefined` here based on server/plugin.ts. Can we make this more typesafe by explicitly calling out `client` could be undefined here? Some comments when it is initialized would be helpful as well, and more details stating that if MDS is enabled, there will be no available client, since we fetch the configured client from the datasource plugin.

server/plugin.ts
```diff
@@ -50,20 +57,28 @@ export class AnomalyDetectionOpenSearchDashboardsPlugin
...
+    let client: ILegacyClusterClient | undefined = undefined;
+
+    if (!dataSourceEnabled) {
+      client = core.opensearch.legacy.createClient('anomaly_detection', {
         plugins: [adPlugin, alertingPlugin],
         customHeaders: { ...customHeaders, ...DEFAULT_HEADERS },
         ...rest,
-      }
-    );
+      });
```
ohltyler, line 77: Can you add comments about how MDS works, and how the client will be undefined when MDS is available since it is configured/fetched from the datasource plugin based on a data source ID? Making the client optional or updating the type to be `| undefined` can make things easier to track and prevent NPEs down the road.

server/routes/ad.ts
```diff
@@ -59,42 +60,85 @@ type PutDetectorParams = {
...
+  
+
+  // post search anomaly results
+  apiRouter.post('/detectors/results/_search/', 
+    adService.searchResults);
+  apiRouter.post('/detectors/results/{dataSourceId}/_search', 
+    adService.searchResults);
+  apiRouter.post('/detectors/results/_search/{resultIndex}/{onlyQueryCustomResultIndex}',
+    adService.searchResults);
+  apiRouter.post('/detectors/results/{dataSourceId}/_search/{resultIndex}/{onlyQueryCustomResultIndex}',
```
ohltyler, line 83: can we be consistent where `dataSourceId` is present in the path? perhaps always earlier/in the beginning, or always at the very end. Suggest to not have in the middle for a few reasons:
- having data source ID before `_search` doesn't make logical sense; the full search path should be the base path before adding optional parameters like datasourceid
- inconsistent, where it is after `results` here, but before `results` on line 97, for example
- more error-prone when constructing the path on client-side to make the calls
- more explicit paths needed. if always at a consistent place in the path, you could have a single utility fn to insert the datasourceid for any given path and greatly reduce all of the explicit ones. As a bonus, if defined under `common/`, it can be used for client and server-side and ensure a lot more consistency

ohltyler (commented):
Great to see this coming together. Have some comments, but nothing too large.

public/pages/Dashboard/Container/DashboardOverview.tsx
```diff
@@ -157,14 +191,24 @@ export function DashboardOverview() {
...
+    const updatedParams = {
+      dataSourceId: MDSOverviewState.selectedDataSourceId,
+    };
+    history.replace({
+      ...location,
+      search: queryString.stringify(updatedParams),
+    })
+    if (dataSourceEnabled ? MDSOverviewState.selectedDataSourceId : true) {
+      intializeDetectors();
+    }
```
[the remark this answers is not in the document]
jackiehanyang: The example you provided will result in initializeDetectors() not being called if MDS is not enabled. When MDS is enabled, selectedDataSourceId should always exist because the data source picker provides a default data source when rendering. If not, the data source picker will display an error banner. Additionally, later in the reducer, I initialized dataSourceId as '' if it doesn't have a truthy value, allowing it to fall back to the local cluster case. Therefore, I think we can remove the if-check here and trust the process we implement later.

opensearch_dashboards.json
```diff
@@ -5,6 +5,7 @@
   "configPath": [
     "anomaly_detection_dashboards"
   ],
+  "optionalPlugins": ["dataSource","dataSourceManagement"],
```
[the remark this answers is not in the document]
jackiehanyang: @BionIT could you please look at this comment? how neo/other plugin teams are handling the NPE case?

public/pages/Dashboard/utils/utils.tsx
```diff
@@ -441,7 +442,8 @@ export const getLatestAnomalyResultsByTimeRange = async (
   anomalySize: number,
   checkLastIndexOnly: boolean,
   resultIndex: string,
-  onlyQueryCustomResultIndex: boolean
+  onlyQueryCustomResultIndex: boolean,
+  dataSourceId = ''
```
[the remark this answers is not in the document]
jackiehanyang: Adding an extra safety net doesn't hurt. I'm just incorporating this optional initializer whenever possible to maintain consistency, and ensuring that when dataSourceId is not present, we can fallback to the local cluster case, so it won't break anything.

public/pages/Dashboard/utils/utils.tsx
```diff
@@ -499,7 +503,8 @@ export const getLatestAnomalyResultsForDetectorsByTimeRange = async (
   detectorNum: number,
   checkLastIndexOnly: boolean,
   resultIndex: string,
-  onlyQueryCustomResultIndex: boolean
+  onlyQueryCustomResultIndex: boolean,
+  dataSourceId = ''
```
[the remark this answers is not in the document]
jackiehanyang: same reply as above

public/pages/Dashboard/utils/utils.tsx
```diff
@@ -605,9 +611,11 @@ export const getAnomalyDistributionForDetectorsByTimeRange = async (
   func: (
     request: any,
     resultIndex: string,
+    dataSourceId: string,
     onlyQueryCustomResultIndex: boolean
   ) => APIAction,
   selectedDetectors: DetectorListItem[],
+  dataSourceId = '',
```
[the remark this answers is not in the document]
jackiehanyang: same reply as above

public/pages/DetectorDetail/containers/DetectorDetail.tsx
```diff
@@ -155,15 +161,17 @@ export const DetectorDetail = (props: DetectorDetailProps) => {
...
-    const getInitialIndices = async () => {
-      await dispatch(getIndices('')).catch((error: any) => {
-        console.error(error);
-        core.notifications.toasts.addDanger('Error getting all indices');
-      });
-    };
-    // only need to check if indices exist after detector finishes loading
-    if (!isLoadingDetector) {
-      getInitialIndices();
+    if (dataSourceEnabled ? dataSourceId : true) {
```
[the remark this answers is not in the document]
jackiehanyang: same reply as above

public/pages/DetectorDetail/hooks/useFetchMonitorInfo.ts
```diff
@@ -18,18 +18,19 @@ import { searchMonitors } from '../../../redux/reducers/alerting';
 
 //A hook which gets AD monitor.
 export const useFetchMonitorInfo = (
-  detectorId: string
+  detectorId: string,
+  dataSourceId = '',
```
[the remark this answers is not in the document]
jackiehanyang: same reply as above

public/pages/DetectorDetail/hooks/useFetchMonitorInfo.ts
```diff
@@ -18,18 +18,19 @@ import { searchMonitors } from '../../../redux/reducers/alerting';
...
   fetchMonitorError: boolean;
   isLoadingMonitor: boolean;
 } => {
   const dispatch = useDispatch();
   useEffect(() => {
-    const fetchAdMonitors = async () => {
-      await dispatch(searchMonitors());
-    };
-    fetchAdMonitors();
+    if (dataSourceEnabled ? dataSourceId : true) {
```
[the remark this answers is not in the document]
jackiehanyang: same reply as above

public/pages/DetectorResults/containers/AnomalyHistory.tsx
```diff
@@ -95,6 +95,7 @@ import {
 import { CoreStart } from '../../../../../../src/core/public';
 import { CoreServicesContext } from '../../../components/CoreServices/CoreServices';
 import { prettifyErrorMessage } from '../../../../server/utils/helpers';
+import { useLocation } from 'react-router-dom';
```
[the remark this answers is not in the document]
jackiehanyang: will run prettier in the next pr :)

public/pages/DetectorResults/containers/AnomalyResults.tsx
```diff
@@ -80,21 +83,28 @@ export function AnomalyResults(props: AnomalyResultsProps) {
   const core = React.useContext(CoreServicesContext) as CoreStart;
   const dispatch = useDispatch();
   const detectorId = props.detectorId;
-  const detector = useSelector(
+  const detector = useSelector(    
     (state: AppState) => state.ad.detectors[detectorId]
   );
+  const dataSourceEnabled = getDataSourcePlugin().dataSourceEnabled;
+  const location = useLocation();
+  const dataSourceId = new URLSearchParams(location.search).get(DATA_SOURCE_ID) || '';
```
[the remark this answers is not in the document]
jackiehanyang: thought about this idea, having a fn to parse url query params like what we do currently on List page. Thought it's fine because right now it's just one line of getting the dataSourceId. But I can make this a utility fn :)

public/pages/DetectorResults/containers/AnomalyResults.tsx
```diff
@@ -80,21 +83,28 @@ export function AnomalyResults(props: AnomalyResultsProps) {
...
 
   useEffect(() => {
     core.chrome.setBreadcrumbs([
       BREADCRUMBS.ANOMALY_DETECTOR,
       BREADCRUMBS.DETECTORS,
       { text: detector ? detector.name : '' },
     ]);
-    dispatch(getDetector(detectorId));
+    if (dataSourceEnabled ? dataSourceId : true) {
```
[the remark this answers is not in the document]
jackiehanyang: same reply as above

public/pages/DetectorResults/containers/AnomalyResults.tsx
```diff
@@ -80,21 +83,28 @@ export function AnomalyResults(props: AnomalyResultsProps) {
...
+      dispatch(getDetector(detectorId, dataSourceId));
+    }
   }, []);
 
   const fetchDetector = async () => {
-    dispatch(getDetector(detectorId));
+    if (dataSourceEnabled ? dataSourceId : true) {
```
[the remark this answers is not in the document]
jackiehanyang: same reply as above

public/pages/DetectorResults/containers/AnomalyResultsLiveChart.tsx
```diff
@@ -130,9 +134,10 @@ export const AnomalyResultsLiveChart = (
           detectionInterval,
           intervals
         );
-        await dispatch(
+        await dispatch(
```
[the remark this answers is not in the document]
jackiehanyang: will run prettier in the next commit :)

public/pages/DetectorsList/containers/List/List.tsx
```diff
@@ -223,12 +240,16 @@ export const DetectorList = (props: ListProps) => {
     });
 
     setIsLoadingFinalDetectors(true);
-    getUpdatedDetectors();
+
+    if (dataSourceEnabled ? state.selectedDataSourceId : true) {
```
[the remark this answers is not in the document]
jackiehanyang: same reply as above

public/pages/DetectorsList/containers/List/List.tsx
```diff
@@ -471,7 +496,9 @@ export const DetectorList = (props: ListProps) => {
         );
       })
       .finally(() => {
-        getUpdatedDetectors();
+        if (dataSourceEnabled ? state.selectedDataSourceId : true) {
```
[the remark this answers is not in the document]
jackiehanyang: same reply as above

public/pages/main/Main.tsx
jackiehanyang: sure, let me update this in the next revision

public/pages/main/Main.tsx
jackiehanyang: I'm not deleting this isLoadingDetectors line, just passing it as a prop from here doesn't make sense. So I moved it inside AnomalyDetectionOverview page and also added another selector from open search reducer as we're also loading sample data/indices - https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/f166bd2463d5d314a6138ee082c8a72c92e26231/public/pages/Overview/containers/AnomalyDetectionOverview.tsx#L79-L80

server/routes/ad.ts
jackiehanyang: I didn't give much thought to where to place dataSourceId in the path since I considered it just a link between the reducer and the server. As long as they aren't conflicting with others, it seemed fine.

I've updated the path to place all dataSourceIds at the end in the latest revision.

server/plugin.ts
jackiehanyang: Neo team provided this initialization method, intended to leave client as undefined as this staged. We conditionally create an  `ILegacyClusterClient` only when MDS is not enabled. If MDS is enabled, no client is created because we would like to create the client later based on the request context when we have a dataSourceId. 

Will add more comments in the latest revision

public/pages/CreateDetectorSteps/hooks/useFetchDetectorInfo.ts
```diff
@@ -43,7 +44,8 @@ export const useFetchDetectorInfo = (
   useEffect(() => {
     const fetchDetector = async () => {
       if (!detector) {
-        await dispatch(getDetector(detectorId));
+        // hardcoding the datasource id for now, will update it later when working on create page
```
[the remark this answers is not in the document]
jackiehanyang: removed

public/pages/DetectorsList/containers/List/List.tsx
```diff
@@ -198,6 +202,9 @@ export const DetectorList = (props: ListProps) => {
     selectedIndices: queryParams.indices
       ? queryParams.indices.split(',')
       : ALL_INDICES,
+    selectedDataSourceId: queryParams.dataSourceId
+      ? queryParams.dataSourceId
+      : '',
```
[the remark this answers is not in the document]
jackiehanyang: prefer using ternary operator here as it covers more falsy value like `0` and `false`

public/pages/Overview/containers/AnomalyDetectionOverview.tsx
```diff
@@ -86,20 +97,12 @@ export function AnomalyDetectionOverview(props: AnomalyDetectionOverviewProps) {
...
-  const getAllSampleIndices = async () => {
-    await dispatch(getIndices(GET_SAMPLE_INDICES_QUERY)).catch((error: any) => {
-      console.error('Error getting all indices: ', error);
-    });
-  };
+  
+  const queryParams = getURLQueryParams(props.location);
+  const [MDSOverviewState, setMDSOverviewState] = useState<MDSOverviewState>({
+    queryParams,
+    selectedDataSourceId: queryParams.dataSourceId ? queryParams.dataSourceId : '',
```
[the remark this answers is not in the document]
jackiehanyang: same as above

public/pages/Dashboard/Container/DashboardOverview.tsx
```diff
@@ -108,6 +128,20 @@ export function DashboardOverview() {
...
+  const handleDataSourceChange = ([event]) => {
+    const dataSourceId = event?.id;
+    if (!dataSourceId) {
+      getNotifications().toasts.addDanger(
+        prettifyErrorMessage('Unable to set data source.')
+      );
+    } else {
+      setMDSOverviewState({
+        queryParams: dataSourceId,
+        selectedDataSourceId: dataSourceId,
```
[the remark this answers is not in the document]
jackiehanyang: my idea is to seperate query param state with selectedDataSourceId state

public/pages/Dashboard/Container/DashboardOverview.tsx
```diff
@@ -204,10 +227,8 @@ export function DashboardOverview(props: OverviewProps) {
     history.replace({
       ...location,
       search: queryString.stringify(updatedParams),
-    })
-    if (dataSourceEnabled ? MDSOverviewState.selectedDataSourceId : true) {
```
huyaboo: Does removing this conditional change the logic of `initializeDetectors()`?

public/pages/DetectorDetail/containers/DetectorDetail.tsx
```diff
@@ -161,17 +177,15 @@ export const DetectorDetail = (props: DetectorDetailProps) => {
   // Getting all visible indices. Will re-fetch if changes to the detector (e.g.,
   // detector starts, result index recreated or user switches tabs to re-fetch detector)
   useEffect(() => {
-    if (dataSourceEnabled ? dataSourceId : true) {
```
huyaboo: Similar comment: does removing this conditional change the logic?

public/pages/DetectorDetail/hooks/useFetchMonitorInfo.ts
```diff
@@ -28,9 +28,7 @@ export const useFetchMonitorInfo = (
 } => {
   const dispatch = useDispatch();
   useEffect(() => {
-    if (dataSourceEnabled ? dataSourceId : true) {
```
huyaboo: Same as above comments

public/pages/Dashboard/Container/DashboardOverview.tsx
jackiehanyang: No, I didn't change any logic inside `initializeDetectors()`, why would it change the logic of it? I explained the reason why removing this conditional check here - https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/722#discussion_r1569214460

huyaboo (approved):

public/pages/Dashboard/Container/DashboardOverview.tsx
```diff
@@ -237,28 +258,31 @@ export function DashboardOverview(props: OverviewProps) {
...
-          onSelectedDataSources: (dataSources) =>
-            handleDataSourceChange(dataSources),
-        }}
-      />
-    );
-  }, [getSavedObjectsClient(), getNotifications(), props.setActionMenu]);
+  let renderDataSourceComponent = null;
+  if (dataSourceEnabled) {
+    const DataSourceMenu =
+      getDataSourceManagementPlugin().ui.getDataSourceMenu<DataSourceSelectableConfig>();
```
ohltyler, line 268: can this trigger an NPE?

ohltyler, line 268: Can this trigger an NPE?

ohltyler: I understand the datasourcenabled check has passed here, but that is from a different plugin. prefer to have this more safely handled

public/pages/DetectorsList/containers/List/List.tsx
```diff
@@ -668,24 +669,27 @@ export const DetectorList = (props: ListProps) => {
...
-          onSelectedDataSources: (dataSources) =>
-            handleDataSourceChange(dataSources),
-        }}
-      />
-    );
-  }, [getSavedObjectsClient(), getNotifications(), props.setActionMenu]);
+  let renderDataSourceComponent = null;
+  if (dataSourceEnabled) {
+    const DataSourceMenu =
+      getDataSourceManagementPlugin().ui.getDataSourceMenu<DataSourceSelectableConfig>();
```
ohltyler, line 675: same comment here. Prefer an NPE check.

public/pages/Overview/containers/AnomalyDetectionOverview.tsx
```diff
@@ -211,26 +237,29 @@ export function AnomalyDetectionOverview(props: AnomalyDetectionOverviewProps) {
...
-          onSelectedDataSources: (dataSources) =>
-            handleDataSourceChange(dataSources),
-        }}
-      />
-    );
-  }, [getSavedObjectsClient, getNotifications, props.setActionMenu]);
+  let renderDataSourceComponent = null;
+  if (dataSourceEnabled) {
+    const DataSourceMenu =
+      getDataSourceManagementPlugin().ui.getDataSourceMenu<DataSourceSelectableConfig>();
```
ohltyler, line 245: same comment

public/pages/main/Main.tsx
```diff
@@ -48,16 +48,34 @@ export function Main(props: MainProps) {
...
+  const dashboardHref = constructHrefWithDataSourceId(
+    APP_PATH.DASHBOARD,
+    '',
+    dataSourceId
+  );
+  const overviewHref = constructHrefWithDataSourceId(
+    APP_PATH.OVERVIEW,
+    '',
+    dataSourceId
+  );
```
ohltyler, line 78: how about just using these directly below instead of creating these extra vars? You already do for detector list page.

public/pages/utils/helpers.ts
ohltyler: Are you going to make this change?

public/pages/Dashboard/utils/utils.tsx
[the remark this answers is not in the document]
ohltyler: Hmm, the problem is it makes the code less readable as it's breaking the type safety. If `dataSourceId` is getting passed in as non-optional, then no check should be done, and so on, all the way up the calling chain, as long as the code is adhering to type safety. That makes it much easier to track the state of props as they are being passed around, as ad-hoc initializations assume there is doubt and uncertainty in the values.

public/pages/DetectorResults/containers/AnomalyResults.tsx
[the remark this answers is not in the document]
ohltyler: Great, can you do this? It will just make everything more maintainable and flexible when we have all of these common processes for handling MDS data. It also ensures consistency that datasourceId is getting initialized to the same value across all pages.

public/pages/main/Main.tsx
ohltyler: Looks like this wasn't updated, could you remove?

public/services.ts
ohltyler: Are you going to make this change?

server/utils/helpers.ts
ohltyler: Are you going to make this change?

server/plugin.ts
ohltyler: Can you add more comments? Don't see any in the latest revision.

server/routes/ad.ts
ohltyler: Cool, sgtm.

public/pages/Dashboard/Container/DashboardOverview.tsx
ohltyler: I understand that's the current mechanism. But given the fact this is technically an optional plugin, let's add a single `?` check to prevent NPE.

public/pages/main/Main.tsx
jackiehanyang: updated in the latest revision

public/pages/Dashboard/Container/DashboardOverview.tsx
jackiehanyang: updated in the latest revision :)

public/services.ts
jackiehanyang: no, this change will be made in UT pr

server/utils/helpers.ts
jackiehanyang: already made the change. And just realized I have already added comment in this function, so no need to add comments

public/pages/DetectorResults/containers/AnomalyResults.tsx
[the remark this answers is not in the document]
jackiehanyang: can I address this in the upcoming pr? Already made the change, but it's in a different branch

server/plugin.ts
jackiehanyang: comment added

public/pages/Dashboard/utils/utils.tsx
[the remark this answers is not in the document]
jackiehanyang: It's not an ad-hoc initialization, I did it on purpose. If dataSourceId is defined as a string and will be used as a string, I don't think initialize it as '' here will break the type safety? It only initializes it when dataSourceId is null, undefined. We're adding a new branch (neo branch) for our application. I tried my best to handle every places where we need to add this branch but I still want to initialization whenever we go into redux/router/server calls to make sure our application won't crash because of any incompatibility.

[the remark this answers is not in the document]
ohltyler: > It only initializes it when dataSourceId is falsy value like null, undefined

On line 43 `dataSourceId` is defined as a `string`, so it can't be `null` or `undefined` if the type is correct. It could only be null or undefined if it's typed as `string | null | undefined`. If it is typed as a string on initialization (I believe it is on line 85 of `DashboardOverview` if I follow the codepath right), then there is no re-initialization needed.

I'm just trying to provide best practices and ways to make the code more readable and typesafe. The more we can explicitly call out what types are, especially on initialization, the easier it is in downstream fns and components that can bind by those typed contracts and not have to worry about edge cases of some values not being what they are expected to be.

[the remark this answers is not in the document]
ohltyler: > We're adding a new branch (neo branch) for our entire application

Also not sure what is meant by this? Also let's keep this open source and using terms like MDS.

public/pages/DetectorResults/containers/AnomalyResults.tsx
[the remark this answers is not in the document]
ohltyler: sure

server/utils/helpers.ts
ohltyler: Can you please make the parameters more typesafe (context, client)? Again it is much easier to follow if we know what `client` is and that it could be undefined. Ideally having a comment about how it could be undefined is helpful. It's just for tracking and readability purposes for the future.

public/pages/DetectorResults/containers/AnomalyHistory.tsx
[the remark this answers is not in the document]
ohltyler: I mention prettier doesn't solve this:
> We're adding a new branch (neo branch) for our entire application

But not a blocker, just a readability thing. Ideally there's a prettier setting that can handle this for us

server/utils/helpers.ts
jackiehanyang: I already did, 3 revisions ago. Please look at the latest code

public/pages/Dashboard/utils/utils.tsx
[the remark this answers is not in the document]
jackiehanyang: removed initialization part in the latest commit. Please take another look thanks

server/utils/helpers.ts
ohltyler: Ah! Missed that with all of the commit filtering. My bad. thanks!

ohltyler (approved):
Thanks for all of the hard work. LGTM for a feature branch merge!

PR #728, opened 2024-04-18
add MDS support for detector creation&Update flow

jackiehanyang:
onboarded MDS for the following API calls by adding datasourceid paths to the redux and node apis:
- apiRouter.post('/detectors/preview/{dataSourceId}', adService.previewDetector);
- apiRouter.get('/detectors/{detectorName}/_match/{dataSourceId}', adService.matchDetector);
- apiRouter.get('/detectors/_count/{dataSourceId}', adService.getDetectorCount);
-  apiRouter.post('/detectors/_validate/{validationType}/{dataSourceId}', adService.validateDetector);
- apiRouter.get('/_mappings/{dataSourceId}', opensearchService.getMapping);

### Revisions:
4/19:
- remove `neo` in variable naming
- create and use enum when referring to list detector page query param
- make universal MDSState and share it between pages

### Description

This pr adds MDS support for detector creation&update flow

https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/assets/41348518/66eb71f2-8a11-4d92-a7cc-fc9f4966fca0

### Issues Resolved
https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/issues/721

### Check List

  - [ ] All tests pass


public/pages/DefineDetector/containers/DefineDetector.tsx
```diff
@@ -218,6 +262,61 @@ export const DefineDetector = (props: DefineDetectorProps) => {
...
+      const DataSourceMenu =
+        getDataSourceManagementPlugin().ui.getDataSourceMenu<DataSourceSelectableConfig>();
+      renderDataSourceComponent = useMemo(() => {
+        return (
+          <DataSourceMenu
+            setMenuMountPoint={props.setActionMenu}
+            componentType={'DataSourceSelectable'}
+            componentConfig={{
+              fullWidth: false,
+              activeOption: [{ id: MDSCreateState.selectedDataSourceId }],
```
huyaboo, line 306: How are we handling the case where `MDSCreateState.selectedDataSourceId = ''` but `Local cluster` is disabled? When this is initially defined, the default value is `''` which means the initial selected option will be the `Local cluster`. I think a potential solution here is to set the initial state of  `MDSCreateState.selectedDataSourceId = undefined` and something like the following:

```Typescript
activeOption: MDSCreateState.selectedDataSourceId === undefined ? [{ id: MDSCreateState.selectedDataSourceId }] : undefined
```

This enables the `DataSourceSelectable` component to handle the default selected choice. Could you please check for other pages as well?

server/models/types.ts
```diff
@@ -110,7 +110,7 @@ export type DetectorResultsQueryParams = {
 };
 
 export type MDSQueryParams = {
-  dataSourceId?: string;
+  dataSourceId: string;
```
huyaboo, line 113: What if `dataSourceId` were falsy here and `Local cluster` is disabled?

huyaboo (commented):
Overall looks good. Please see comments regarding how your page is handling `DataSourceSelectable` components and check if `activeOption` is being handled correctly in the case that no `dataSourceId` is provided.

ohltyler (commented):
This is great! Nothing big, few minor comments. Besides that, could you also add a lot more details in the PR description? It is really helpful for tracking and auditing of changes to have at the very least some details in each PR regarding the impact of the change, the implementation changes made (e.g., 'onboarded MDS for the following x API calls by adding datasourceid paths to the redux and node apis', etc. In addition to what is currently missing if it's an ongoing process, what future PRs will finish x,y,z, and referencing to the meta issue. Especially when it comes to hunting down bugs or user-reported issues.

public/components/CreateDetectorButtons/CreateDetectorButtons.tsx
```diff
@@ -12,14 +12,24 @@
 import { EuiFlexGroup, EuiFlexItem, EuiButton } from '@elastic/eui';
 import React from 'react';
 import { APP_PATH, PLUGIN_NAME } from '../../utils/constants';
+import { useLocation } from 'react-router-dom';
+import { constructHrefWithDataSourceId, getDataSourceFromURL } from '../../pages/utils/helpers';
 
 export const CreateDetectorButtons = () => {
+  const location = useLocation();
+  const neoQueryParams = getDataSourceFromURL(location);
```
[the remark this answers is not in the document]
jackiehanyang: will update it in a new revision

public/pages/DefineDetector/components/Timestamp/Timestamp.tsx
```diff
@@ -18,16 +18,23 @@ import ContentPanel from '../../../../components/ContentPanel/ContentPanel';
...
 
 interface TimestampProps {
   formikProps: FormikProps<DetectorDefinitionFormikValues>;
 }
 
 export function Timestamp(props: TimestampProps) {
   const dispatch = useDispatch();
+  const location = useLocation();
+  const neoQueryParams = getDataSourceFromURL(location);
+  const dataSourceId = neoQueryParams.dataSourceId;
```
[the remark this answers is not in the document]
jackiehanyang: I could make it just online like 
```
  const dataSourceId = getDataSourceFromURL(useLocation()).dataSourceId;
```
but can I leave it as this way? because:
1. is we need to location object elsewhere, we don't need to define it again
2. when MDS is going to have more field in the future, it's more organized to update MDSQueryParams interface and introduce those new fields here

public/pages/DefineDetector/containers/DefineDetector.tsx
```diff
@@ -50,28 +56,57 @@ import { Detector, FILTER_TYPES } from '../../../models/interfaces';
...
+import queryString from 'querystring';
 
 interface DefineDetectorRouterProps {
   detectorId?: string;
 }
 
+interface MDSCreateState {
+  queryParams: MDSQueryParams;
+  selectedDataSourceId: string;
+}
```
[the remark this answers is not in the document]
jackiehanyang: Will update in the new revision

public/pages/DefineDetector/containers/DefineDetector.tsx
```diff
@@ -218,6 +262,61 @@ export const DefineDetector = (props: DefineDetectorProps) => {
     }
   };
 
+  const handleDataSourceChange = ([event]) => {
+    const dataSourceId = event?.id;
+    if (!dataSourceId) {
+      getNotifications().toasts.addDanger(
+        prettifyErrorMessage('Unable to set data source.')
+      );
```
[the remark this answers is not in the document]
jackiehanyang: not so sure... MDS team provided this usage in their component guideline doc

public/pages/utils/helpers.ts
```diff
@@ -134,5 +139,45 @@ export const getSampleDetectorsQueryParamsWithDataSouceId = (
...
+
+  // Set up base parameters for '/detectors'
+  if (basePath.startsWith('/detectors')) {
+    url.set('from', '0');
+    url.set('size', '20');
+    url.set('search', '');
+    url.set('indices', '');
+    url.set('sortField', 'name');
+    url.set('sortDirection', 'asc');
+  }
```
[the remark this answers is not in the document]
jackiehanyang: will update in the new revision

server/models/types.ts
jackiehanyang: as I replied above, I think we have talked about this before and the decision is data source picker will throw error banner in that case

public/pages/DefineDetector/containers/DefineDetector.tsx
huyaboo: The case I'm referring to is how you set the initial render of the page. Currently, if `Local cluster` is disabled and you navigate to the `/app/anomaly-detection-dashboards#/`, it will throw the toast message every time which should be avoided. This is because `MDSCreateState.selectedDataSourceId = ''` when this is not valid.

From datasource UI component side, it is not possible for a user to select `Local cluster` because the component will check to see if `Local cluster` is disabled and add/omit it if necessary. Of course, the toast should still be thrown if the user intentionally/unintentionally sets the url to `dataSourceId=` because it's invalid.

jackiehanyang: ic, this is the issue that AD and IMS have in the future playground, right? Let's address it in the next pr, I haven't pull latest change from OSD in this pr. So the bug has not been exposed in my local yet

huyaboo: From MDS side, here should be the expected behaviors:

If `Local cluster` is enabled:
- The initial render of the page should have `defaultDataSource` as the selected option, not the `Local cluster`. This can be easily done if `activeOptions = undefined`. The url should reflect this choice with `?dataSourceId=<defaultDataSource>`
- The user should be able to select `Local cluster` without any Toast message being shown (and the url should reflect this choice with `?dataSourceId=`)
- The `?dataSourceId=` should be a valid url and this should represent the `Local cluster` being selected

If `Local cluster` is disabled:
- The initial render of the page should have `defaultDataSource` as the selected option, not the `Local cluster`. This can be easily done if `activeOptions = undefined`. The url should reflect this choice with `?dataSourceId=<defaultDataSource>`
- The user shouldn't be able to select `Local cluster` from UI component (this is datasource UI responsibility; this should already be covered)
- The `?dataSourceId=` can either do one of 2 things:
    - Throw the toast message because the `Local cluster` is attempted to be made but is not valid
    - **(Preferred)**: redirect the url to the `?dataSourceId=<defaultDataSource>` and put the `defaultDatasource` as the selected option. If this is non-trivial, this is the preferred option

jackiehanyang: thanks for the details! I will do a sanity check on all the data source picker components and make adjustment in the next pr. Wanted to focus on adding MDS support for creation&update flow in this pr

huyaboo: @kgcreative I may need confirmation if this makes sense from UX perspective. For right now, on datasource UI side, `defaultDataSource` is prioritized over `Local cluster` so we shouldn't change these priorities.

public/pages/utils/helpers.ts
```diff
@@ -160,12 +160,12 @@ export const constructHrefWithDataSourceId = (
...
-    url.set('search', '');
-    url.set('indices', '');
-    url.set('sortField', 'name');
-    url.set('sortDirection', 'asc');
+    url.set(DETECTORS_QUERY_PARAMS.FROM, '0');
+    url.set(DETECTORS_QUERY_PARAMS.SIZE, '20');
+    url.set(DETECTORS_QUERY_PARAMS.SEARCH, '');
+    url.set(DETECTORS_QUERY_PARAMS.INDICES, '');
+    url.set(DETECTORS_QUERY_PARAMS.SORT_FIELD, DETECTORS_QUERY_PARAMS.NAME);
+    url.set(DETECTORS_QUERY_PARAMS.SORT_DIRECTION, SORT_DIRECTION.ASC);
```
ohltyler, line 168: You can populate all of the values using `DEFAULT_QUERY_PARAMS in the same constants file.

server/utils/constants.ts
```diff
@@ -46,6 +46,16 @@ export enum SORT_DIRECTION {
...
 
+export enum DETECTORS_QUERY_PARAMS {
+  FROM = 'from',
+  SIZE = 'size',
+  SEARCH = 'search',
+  INDICES = 'indices',
+  SORT_FIELD = 'sortField',
+  SORT_DIRECTION = 'sortDirection',
+  NAME = 'name',
+}
```
ohltyler, line 57: Rather than defining in multiple places: we actually can create shared constants if we add a top-level `common/` dir that's accessible by both client and server-side. Could you add this and at least add the constants file that contains this same data (+ the defaults)? See [flow framework's common/index.ts file](https://github.com/opensearch-project/dashboards-flow-framework/blob/main/common/index.ts) for reference. 

We don't need to move them all now; but we can at least stop adding to the tech debt here.

ohltyler (commented):
Finished another round, getting close. I'll wait for the open threads to be resolved as well. Can you please update the PR description as well?

huyaboo (approved):
Approved with 2 action items:
- Address remaining comments
- After merge, align with the expected behaviors for the `DataSourceSelectable` components and how to handle url cases gracefully

server/utils/constants.ts
jackiehanyang: this is not defined multiple times? We have been using string instead of enums in our package so far. And since you commented earlier about using enum for this part, so I created these name enums.

public/pages/utils/helpers.ts
jackiehanyang: sure let me update it

server/utils/constants.ts
ohltyler: Right, I meant adding this + the defaults in a constants file. No blocker, just a suggestion so we can use this + other defined constants across both public and server side.

public/pages/DefineDetector/containers/DefineDetector.tsx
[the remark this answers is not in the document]
huyaboo: The callback function will either return `[]` (error state), `[{id: '', label: 'Local cluster']` (if `Local cluster` is enabled) or `[{id: <dataSourceId>, label: <dataSourceTitle>}]`. Thus, `event` can be undefined and it's expected that Plugins handle this case as an invalid/error state. 

One example of `[]` being returned is when you pass in an `activeOption` and you specify a `dataSourceFilter` that filters out the datasource. Since the `dataSourceFilter` function removes the `activeOption` as a valid choice, the component will return `[]`.

ohltyler (approved):
LGTM! Lets track the open discussions.

PR #732, opened 2024-04-23
update actionOption attribution for all data soure components

jackiehanyang:
### Description

Previously, we define dataSourceId as empty if it doesn't exist url. However, mds team introduced a change that treats empty dataSourceId as the local cluster scenario, and only undefined actionOption will render the default remote data source. Therefore, we have to adopt this change and assign dataSourceId as undefined when it's not presented in the url.

### Issues Resolved

https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/issues/721

### Check List

  - [ ] All tests pass

### DIFF
--- a/public/models/interfaces.ts
+++ b/public/models/interfaces.ts
@@ -330,5 +330,5 @@ export interface ValidationSettingResponse {
 
 export interface MDSStates {
   queryParams: MDSQueryParams;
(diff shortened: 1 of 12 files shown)


owaiskazi19 (approved):
MDS should handle the type correctly here and we shouldn't make it as undefined.

PR #737, opened 2024-04-29
MDS local cluster scanrio bug fixes

jackiehanyang:
### Description

This PR includes the following bug fixes:
- side navigation bar link doesn't work when data source is local cluster
- when local cluster is selected, it always jump back to default data source selection on fresh 
- the first breadcrumb doesn't work on list page when local cluster is selected

The above two bugs are fixed by making the following changes:
- update breadcrumbs setting on list page. The previous implementation inverted MDS enabled case and MDS disabled case
- make sure when checking if dataSourceId is set in the url, only set dataSourceId as undefined when the query param dataSourceId is undefined
- introduce a new prop - landingDataSourceId that keeps track of if dataSourceId exist in the url on landing pages (Overview page and Dashboard page). In this way, default data source will be selected when opening AD from OSD side navigation bar but select the data source in the url when fresh the page. 

All bugs can be reproduced on future playground side, and here's a screen record that shows the result after this bug fixes. The screen record was tested with turning on/off `data_source.enabled` setting and `data_source.hideLocalCluster` setting

https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/assets/41348518/a8047e83-d3ce-4050-8a7a-2ff4b469f54b

### Issues Resolved

### Check List

  - [ ] All tests pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,13 +25,13 @@
   "devDependencies": {
     "@testing-library/user-event": "^12.1.6",
     "@types/react-plotly.js": "^2.6.0",
-    "@types/redux-mock-store": "^1.0.1",
+    "@types/redux-mock-store": "^1.0.6",
     "babel-polyfill": "^6.26.0",
     "eslint-plugin-no-unsanitized": "^3.0.2",
     "eslint-plugin-prefer-object-spread": "^1.2.1",
     "lint-staged": "^9.2.0",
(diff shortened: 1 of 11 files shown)


huyaboo (approved):
LGTM

amitgalitz (approved):

PR #757, opened 2024-05-15
Fix handling of special characters in categorical values

kaituo:
### Description

This PR addresses a bug that occurred when categorical values contained special characters such as colons. Previously, entities were stored as a comma-separated name-value pair string in the heatmap cell's custom data field. When a cell was clicked, the custom data was retrieved and the string was split via colons. This caused errors if the categorical value itself contained colons or if multiple name-value pairs included commas.

To fix this issue, this PR changes the storage format to a JSON string in the cell's custom data. When a cell is clicked, the JSON object is restored. Since JSON naturally handles special characters, this eliminates the need for additional special character handling.

Additionally, the previous implementation stored a newline-separated name-value pair in the y-axis and a comma-separated name-value pair in the cell tooltip. With entities now stored as JSON in custom data, we no longer have the flexibility to store these different formats. This PR standardizes the format to newline-separated name-value pairs for both the y-axis and cell tooltip.

Testing Done:
* Conducted end-to-end testing in preview, historical, and real-time use cases.
* Added unit tests.

### Check List

  - [X] All tests pass

### DIFF
--- a/public/pages/AnomalyCharts/utils/__tests__/anomalyChartUtils.test.ts
+++ b/public/pages/AnomalyCharts/utils/__tests__/anomalyChartUtils.test.ts
@@ -15,6 +15,8 @@ import {
   getAnomalySummary,
   convertAlerts,
   generateAlertAnnotations,
+  buildHeatmapPlotData,
+  ANOMALY_HEATMAP_COLORSCALE,
 } from '../anomalyChartUtils';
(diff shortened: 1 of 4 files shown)


kaituo:
Not sure why coverage CI failed as I only added tests and didn't delete tests. Will leave it as it is.

jackiehanyang:
Thanks for the fix and adding tests. LGTM!

jackiehanyang (approved):

PR #772, opened 2024-06-07
Update Frontend for Custom Result Index Query and Fix Issues

kaituo:
### Description

This PR finalizes the frontend changes related to [PR #1225](https://github.com/opensearch-project/anomaly-detection/pull/1225). The custom result index query now uses an index pattern instead of a single index.

Additionally, this PR addresses an issue where missing custom result indices would appear because the original code checked for the existence of an index name, but now we use it as a prefix. We have updated the logic to perform a prefix search instead of checking for index name equality.

This PR also resolves issue #765 by downgrading the version of jest-canvas-mock.

Testing Done:
* Added unit tests.
* Verified that the custom result index missing callout is not shown.
* Confirmed that the frontend can still display old and new results after a rollover.

### Issues Resolved

https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/issues/765

### Check List

  - [X] All tests pass

### DIFF
--- a/public/pages/DetectorDetail/containers/DetectorDetail.tsx
+++ b/public/pages/DetectorDetail/containers/DetectorDetail.tsx
@@ -45,7 +45,7 @@ import {
   getDetector,
   stopHistoricalDetector,
 } from '../../../redux/reducers/ad';
-import { getIndices } from '../../../redux/reducers/opensearch';
+import { getAliases, getIndices } from '../../../redux/reducers/opensearch';
 import { getErrorMessage, Listener } from '../../../utils/utils';
 import { darkModeEnabled } from '../../../utils/opensearchDashboardsUtils';
 import { BREADCRUMBS, MDS_BREADCRUMBS } from '../../../utils/constants';
@@ -145,7 +145,7 @@ export const DetectorDetail = (props: DetectorDetailProps) => {
   - If the detector is still loading, the result index is not missing.
   - If the result index retrieved from the detector is empty, it is not missing.
   - If cat indices are being requested, the result index is not missing.
-  - If visible indices are empty, it is likely there is an issue retrieving existing indices.
+  - If visible indices/aliaes are empty, it is likely there is an issue retrieving existing indices.
     To be safe, we'd rather not show the error message and consider the result index not missing.
   - If the result index is not found in the visible indices, then it is missing.
   */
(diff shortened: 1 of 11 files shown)


public/pages/DetectorDetail/utils/helpers.tsx
```diff
@@ -142,11 +142,22 @@ export const getDetectorStateDetails = (
...
+ * @param index - The prefix string to check against the `index` properties of the `CatIndex` objects.
+ * @param indices - An array of `CatIndex` objects to search through.
+ * @returns A boolean value indicating whether any `CatIndex` object's `index` property starts with the specified prefix.
+ */
 export const containsIndex = (index: string, indices: CatIndex[]) => {
   let containsIndex = false;
   if (!isEmpty(indices)) {
     indices.forEach((catIndex: CatIndex) => {
-      if (get(catIndex, 'index', '') == index) {
+      if (get(catIndex, 'index', '').startsWith(index)) {
```
amitgalitz, line 160: Could this give an issue where another unrelated index exists that has the same prefix with other additional characters but its not used for AD

amitgalitz (approved):

public/pages/DetectorDetail/utils/helpers.tsx
kaituo: Yes, it is possible. For example, customer configured opensearch-ad-result-a, but customer has another index opensearch-ad-result-ab.  Index opensearch-ad-result-ab-history-0607-1 exists, but opensearch-ad-result-a-history-0607-1 does not. I will change to check for alias.

kaituo:
In the latest commit, I have made the following changes:

- **Rebased Code:** Rebased my code from the main branch.
- **Alias Check:** Updated the alias check to use equality instead of a prefix check. For more details, refer to this [discussion](https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/772#discussion_r1631790152).
- **Removed ISM Plugin in Custom Index:** Removed the use of the ISM plugin in the custom index part of the create detector workflow. The `min_age` and similar parameters are required and cannot be omitted. Even if these parameters are made optional, they default to empty strings, which the backend AnomalyDetector' parse method expects to be integers.
    Screenshot 2024-06-08 at 11 36 09 AM
- **Configuration Review Page:** Modified the configuration review page to exclude ISM parameter units when the custom result index is empty for a cleaner appearance.
  - Before:
    Screenshot 2024-06-08 at 11 18 17 AM
  - After:
    Screenshot 2024-06-08 at 11 29 45 AM

public/pages/DetectorDetail/containers/DetectorDetail.tsx
```diff
@@ -197,20 +202,21 @@ export const DetectorDetail = (props: DetectorDetailProps) => {
...
   // detector starts, result index recreated or user switches tabs to re-fetch detector)
   useEffect(() => {
-    const getInitialIndices = async () => {
+    const getInitialIndicesAliases = async () => {
       try {
         await dispatch(getIndices('', dataSourceId));
+        await dispatch(getAliases('', dataSourceId));
       } catch (error) {
         console.error(error);
         core.notifications.toasts.addDanger('Error getting all indices');
```
amitgalitz, line 214: nit: Error getting all indices and aliases?

public/pages/DetectorDetail/containers/DetectorDetail.tsx
```diff
@@ -197,20 +202,21 @@ export const DetectorDetail = (props: DetectorDetailProps) => {
...
 
-  // Getting all visible indices. Will re-fetch if changes to the detector (e.g.,
+  // Getting all visible indices & aliases. Will re-fetch if changes to the detector (e.g.,
   // detector starts, result index recreated or user switches tabs to re-fetch detector)
   useEffect(() => {
-    const getInitialIndices = async () => {
+    const getInitialIndicesAliases = async () => {
       try {
         await dispatch(getIndices('', dataSourceId));
+        await dispatch(getAliases('', dataSourceId));
```
amitgalitz, line 211: Just wondering when you make two call like these, we are just populating the redux store one by one so we have them available?

amitgalitz (approved):
LGTM, thanks for fixing the alias part

public/pages/DetectorDetail/containers/DetectorDetail.tsx
kaituo: yes, changed to core.notifications.toasts.addDanger('Error getting all indices or aliases');

public/pages/DetectorDetail/containers/DetectorDetail.tsx
kaituo: yes

amitgalitz (approved):

PR #854, opened 2024-08-29
Adding remote indices and multi index functionality

amitgalitz:
### Description

1. Adding the ability to fetch information on remote clusters and select which clusters we want to choose datasources from
2. Adds the ability to get information from remote indices such as the mapping
3. Add the ability to select more the one index/alias/index-pattern
4. Only resets the chosen model configuration if an index has been removed but not added

### Check List

  - [x] All tests pass


public/pages/DefineDetector/components/Datasource/DataSource.tsx
```diff
@@ -99,16 +195,45 @@ export function DataSource(props: DataSourceProps) {
...
 
-  const isDifferentIndex = () => {
-    return props.isEdit && indexName !== props.origIndex;
+  const isSelectedOptionIndexRemoved = (
+    newSelectedOptions: { label: string }[] = indexNames,
+    oldSelectedOptions: { label: string }[] = props.formikProps.values.index
+  ) => {
+    const newSelectedOptionsSet = new Set(newSelectedOptions);
+    const indexRemoved: boolean =
+      !(newSelectedOptions && oldSelectedOptions) ||
```
kaituo, line 204: If both newSelectedOptions and oldSelectedOptions are null or undefined, would it make more sense for isSelectedOptionIndexRemoved to return false instead of true, because if both are null or undefined, there is no change in the selection state — technically, no option was "removed." ?

amitgalitz: Originally if both were null it was okay for me to set it to true because then it would also make sense to reset the model configuration if we can't find any selected indices. 
I think actually though we pretty much never hit the case where both are undefined so it wasn't making too much of an effect and it logically makes more sense to say indexRemoved is false in that case and if we got to this point we would have had an indexRemoved previously.
I tested it out and removing `!(newSelectedOptins && oldSelectedOptions)` and adding:
```
    if (_.isEmpty(oldSelectedOptions) && _.isEmpty(newSelectedOptions)) {
      return false;
    }
```
also works

kaituo: make sense

kaituo:
cypress test is failing. Can you either fix it or make sure it passes locally?

amitgalitz:
> cypress test is failing. Can you either fix it or make sure it passes locally?

It does pass locally, I'm not sure why its failing here.

kaituo (approved):

kaituo:
Cypress test failed with a different reason now:

plugins/anomaly-detection-dashboard      03:56        4        3        1        -        - │
  │    s-plugin/vis_augmenter/view_anomaly                                                         │
  │    _events_spec.js

kaituo:
I think I fixed the timeout error during historical cypress tests in https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/859 and https://github.com/opensearch-project/opensearch-dashboards-functional-test/pull/1537

Not sure about events_spec.js cypress failure above.

kaituo (approved):

server/routes/opensearch.ts
```diff
@@ -350,12 +405,34 @@ export default class OpenSearchService {
...
+      // make call to fields_caps
+      if (remoteIndices.length) {
+        const fieldCapsResponse = await callWithRequest('transport.request', {
+          method: 'GET',
+          path:
+            remoteIndices.toString() + '/_field_caps?fields=*&include_unmapped',
         });
+        remoteMappings = convertFieldCapsToMappingStructure(fieldCapsResponse);
+      }
+      Object.assign(mappings, remoteMappings);
```
jackiehanyang, line 432: Does this Object.assign works fine on merging nested structure? Should we create a new map object instead of mutating the `mappings` variable directly?

jackiehanyang (approved):

jackiehanyang:
overall lgtm, just I see E2E CI is failing. Please look into that to make sure it won't be a RC build blocker

server/routes/opensearch.ts
amitgalitz: I was reading online that Object.assign is more efficient which can be important as we can have indices with very large mappings. I tested it on complicated nested structures

PR #889, opened 2024-10-14
Fix issues in running historical analysis and custom result index section

jackiehanyang:
### Description

Fix the following issues:

1. wrong route path in running historical analysis
2. when disabling customer result index after enabling it, the field is not getting reset to undefined which leads an error on create detector review page says empty custom result index name

### Issues Resolved

### Check List

  - [x] All tests pass


public/redux/reducers/ad.ts
```diff
@@ -492,10 +492,8 @@ export const startHistoricalDetector = (
   startTime: number,
   endTime: number
 ): APIAction => {
-  const baseUrl = `${AD_NODE_API.DETECTOR}/${detectorId}`;
-  const url = dataSourceId
-    ? `${baseUrl}/${dataSourceId}/start`
-    : `${baseUrl}/start`;
+  const baseUrl = `${AD_NODE_API.DETECTOR}/${detectorId}/start`;
```
amitgalitz, line 495: should we have `_start` instead of `start`? or both can be used?
ref: https://opensearch.org/docs/latest/observing-your-data/ad/api/#start-detector-job

jackiehanyang: we have to use `start` here. It's different from the `_start` in the doc you pasted

jackiehanyang:
e2e CI failed for this error, not related to this pr:
```
Run if timeout 600 grep -q "bundles compiled successfully after" <(tail -n0 -f dashboard.log); then
Timeout for 600 seconds reached. OpenSearch Dashboards did not finish compiling.
```

Did manual e2e verification

PR #1064, opened 2025-07-24
fix ribon encoding issue in contextual launch

jackiehanyang:
### Description

[Describe what this change achieves]

### Issues Resolved

### Check List

  - [ ] All tests pass

### DIFF
--- a/public/pages/DetectorResults/containers/AnomalyResultsTable.tsx
+++ b/public/pages/DetectorResults/containers/AnomalyResultsTable.tsx
@@ -34,6 +34,8 @@ import { CoreStart } from '../../../../../../src/core/public';
 import { CoreServicesContext } from '../../../components/CoreServices/CoreServices';
 import { useLocation } from 'react-router-dom';
 import { getDataSourceFromURL } from '../../../../public/pages/utils/helpers';
+import { setStateToOsdUrl } from '../../../../../../src/plugins/opensearch_dashboards_utils/public';
+import { opensearchFilters, IIndexPattern } from '../../../../../../src/plugins/data/public';
 
 //@ts-ignore
 const EuiBasicTable = EuiBasicTableComponent as any;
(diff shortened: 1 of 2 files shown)


public/pages/DetectorResults/containers/AnomalyResultsTable.tsx
```diff
@@ -78,6 +78,14 @@ export function AnomalyResultsTable(props: AnomalyResultsTableProps) {
       ? props.anomalies.filter((anomaly) => anomaly.anomalyGrade > 0)
       : [];
 
+        // Helper function to properly quote values for rison encoding
+  const quoteRisonValue = (value: string): string => {
+    if (value.includes(' ') || value.includes(',') || value.includes('(') || value.includes(')') || value.includes('!') || value.includes(':') || value.includes("'")) {
```
kaituo, line 83: Should we also include characters such as `*`, `\t`, `\r`, `"` or `%` that can break the URL‑encoded rison string?  Can we rely on `rison.encode()` (or `encode_rison_component`) instead of manual quoting?

jackiehanyang: Those are not considered as special characters. Only the ones I included in the helper function will cause encoding issue

amitgalitz (approved):

public/pages/DetectorResults/containers/AnomalyResultsTable.tsx
kaituo: I tried to use 

{"value": 10.24, "timestamp": "2020-01-01 00:01:00", "host": "O'Brien"}

here is what I got

Screenshot 2025-07-24 at 5 11 40 PM

jackiehanyang: I was reading a wrong issue about ribon encoding, it only listed out the characters that I initially included. Updated the implementation to use the `setStateToOsdUrl` provided by OSD to build the url - https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/opensearch_dashboards_utils/public/state_management/url/osd_url_storage.ts#L95

PR #1120, opened 2025-11-17
adding data selector for index management

amitgalitz:
### Description

adding data selector to indices list page

### Check List


public/pages/DailyInsights/components/IndicesManagement.tsx
```diff
@@ -124,34 +181,6 @@ const fetchInsightsStatus = async () => {
   }
 };
 
-  const handleDataSourceChange = (dataSources: any[]) => {
```
jackiehanyang: nit: for line 174 and 175 above, why doing it in http call way instead of calling it through router?

PR #1189, opened 2026-05-04
feat: disable unsupported features on multi-tenant services data sources

jackiehanyang:
## Description

Disables three AD features on multi-tenant services collections as part of the AD multi-tenant services onboarding. These are UI/frontend gates driven by the data source's engine type — backend route guarding and workspace ACL will come in a follow-up PR that mirrors the pattern from [alerting-dashboards-plugin#1415](https://github.com/opensearch-project/alerting-dashboards-plugin/pull/1415).

## Features disabled on AOSS

| Feature | Why disabled |
|---|---|
| Default result index | multi-tenant services data plane only supports custom result indices scoped to the `opensearch-ad-plugin-result-*` prefix. |
| Flattened custom result index | AOSS ingest pipelines do not support the `script` processor, which the flatten pipeline depends on. |
| Historical analysis | Not in scope for multi-tenant services V2 P0. |

### New `public/utils/dataSourceUtils.ts`

Exports `isServerlessDataSource(dataSourceId?)`. Reads the `data-source` saved object and returns `true` when `attributes.dataSourceEngineType === 'OpenSearch Serverless'`. Returns `false` for the local cluster (no `dataSourceId`) or when the lookup fails.

Engine type string matches core: [`src/plugins/data_source/common/data_sources/types.ts`](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data_source/common/data_sources/types.ts) (`DataSourceEngineType.OpenSearchServerless = 'OpenSearch Serverless'`).

### Custom result index (`CustomResultIndex.tsx`, `DefineDetector.tsx`)

When `isServerless === true`:
- The "Enable custom result index" checkbox is replaced by an info callout: *"Custom result index is required on OpenSearch Serverless."*
- Internal `enabled` state is forced to `true` so the index-name field, lifecycle-management toggle, and min-age/size/TTL fields still render.
- The "Enable flattened custom result index" block is hidden entirely.
- `formikValues.flattenCustomResultIndex` is forced to `false` via `setFieldValue` so the detector payload never carries an enabled flag.

`DefineDetector.tsx` resolves `isServerless` once per selected data source id via effect and passes it down.

### Historical analysis (`DetectorDetail.tsx`, `DetectorJobs.tsx`, `AnomalyResults.tsx`)

- `DetectorDetail.tsx`: filters out the "Historical analysis" tab, and does not register the `/detectors/:id/historical` Route (so a direct URL paste does not render `HistoricalDetectorResults`). Resolves `isServerless` via effect.
- `DetectorJobs.tsx` (create-detector step 3): hides the `HistoricalJob` form block and forces `historical: false` if the value was previously set.
- `AnomalyResults.tsx`: `isServerless` added to the props interface (plumbing only — the existing `onSwitchToHistorical` prop is currently unreferenced inside the component; gating downstream CTAs will land with the same value when those CTAs are added back).

Not touched in this PR: the `DetectorsList` "Historical analysis" column (read-only status for existing detectors, cannot start a new one).

## Out of scope

- Backend route gating / workspace ACL — follow-up PR mirroring alerting-dashboards-plugin#1415.
- `opensearch` → `multi-tenant services` client migration — lives in Neo (BFF), not this repo.
- Security header contract — lives in Neo.
- Regional endpoint routing — lives in Neo / OSD core datasource plumbing.
- Core search result reuse (no AD search result API) — separate PR.

## Testing

- `../../node_modules/.bin/tsc --noEmit -p tsconfig.json` — 0 errors from this change. (Pre-existing `@xyflow/react` lib-level noise is unrelated and present on `main`.)
- `yarn test:jest --watchAll=false` — **909 tests passed / 88 snapshots / 0 failures** in 60s.
- Targeted runs:
  - `DefineDetector.test.tsx` — 6/6 pass, 4 snapshots match.
  - `DetectorDetail`/`DetectorJobs`/`HistoricalJob`/`AnomalyResults` — 34 pass, 1 pre-existing skipped, 10 snapshots match.

Manual verification not performed (no AOSS collection available in a local dev stack); the branch is purely additive and defaults to the pre-existing behavior when `isServerless === false`.

## Check List

- [ ] New functionality includes testing. — frontend gates, no new logic paths; existing tests cover the default (`isServerless === false`) branch. Targeted multi-tenant services-branch tests can be added in follow-up along with an `isServerlessDataSource` unit test.
- [x] New functionality has been documented in this PR description.
- [x] Commits follow Conventional Commits.

### DIFF
--- a/public/pages/DefineDetector/components/CustomResultIndex/CustomResultIndex.tsx
+++ b/public/pages/DefineDetector/components/CustomResultIndex/CustomResultIndex.tsx
@@ -43,10 +43,17 @@ interface CustomResultIndexProps {
   useDefaultResultIndex?: boolean;
   resultIndex?: string;
   formikProps: FormikProps<DetectorDefinitionFormikValues>;
+  // When true, the detector targets an OpenSearch Serverless (AOSS) collection.
+  // Serverless mandates a custom result index (no default index is created) and
+  // does not support the flattened-result-index ingest pipeline (script
+  // processor is not available). We hide the toggles and hard-wire the values.
+  isServerless?: boolean;
 }
 
 function CustomResultIndex(props: CustomResultIndexProps) {
-  const [enabled, setEnabled] = useState<boolean>(!!props.resultIndex);
+  const { isServerless = false } = props;
+  // On serverless, "Enable custom result index" is implicit — no toggle, always on.
+  const [enabled, setEnabled] = useState<boolean>(isServerless || !!props.resultIndex);
   const [customResultIndexConditionsEnabled, setCustomResultIndexConditionsEnabled] = useState<boolean>(true);
   const customResultIndexMinAge = get(props.formikProps, 'values.resultIndexMinAge');
   const customResultIndexMinSize = get(props.formikProps, 'values.resultIndexMinSize');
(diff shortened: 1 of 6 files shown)


jackiehanyang:
codecov/project ci failure is unrelated to this pr

kaituo (approved):

PR #1226, opened 2026-07-17
Migrate Jest test suite to Jest 30 + jsdom 26

### LINKED ISSUE #1225: Migrate plugin tests to Jest 30 + jsdom 26
# Migrate plugin tests to Jest 30 + jsdom 26

## Background
Core PR opensearch-project/OpenSearch-Dashboards#12381 upgrades OpenSearch-Dashboards to
**Jest 30.4.2** + **jsdom 26.1.0** (from Jest 28 / jsdom 16). Bundled plugins are separate
git repos but run their tests against the **parent repo's** `node_modules/.bin/jest`, so they
break once that PR merges. This issue tracks migrating this plugin. See
opensearch-project/OpenSearch-Dashboards#12381 for the reference implementation of every
change category.

## Goal
`yarn test:jest` passes green under Jest 30.4.2 + jsdom 26.1.0, with no content regressions.

## Setup to verify locally
1. In the **parent** repo: check out the opensearch-project/OpenSearch-Dashboards#12381 branch
   and `yarn osd bootstrap`.
   - ⚠️ Bootstrap exits 1 on a plugin cross-version consistency check (formik/plotly
     version mismatches) — this is NON-fatal. Confirm Jest installed:
     `./node_modules/.bin/jest --version` → 30.x.
2. In the plugin repo: create a migration branch. Run the suite with `--ci --maxWorkers=2`
   (shared box) to get a baseline, then work through the checklist.

---

### Config (`test/jest.config.js`)
- [ ] Add `jest-location-mock` as the **first** entry of `setupFilesAfterEnv`.
- [ ] Add `testEnvironmentOptions: { url: 'http://localhost:5601' }`.
      ⚠️ This changes `window.location.origin` to `http://localhost:5601` — update any test
      that hardcodes the old `http://localhost` origin.
- [ ] Add `snapshotFormat: { escapeString: true, printBasicPrototype: true }`
      (Jest 29 flipped these defaults; the shim keeps existing snapshots valid).
- [ ] If config **extends the shared base** (`...config` from `src/dev/jest/config`): do NOT
      override inherited array keys (`setupFilesAfterEnv`, `moduleNameMapper`,
      `transformIgnorePatterns`) — spread the base back in. Extending configs need far less work.

### Setup file(s)
- [ ] Make any `matchMedia` mock `configurable: true`.
- [ ] Set `process.env.HOST = 'http://localhost:5601'` — **required, and load-bearing.** Once
      jest-location-mock is active it re-creates `window.location` from `process.env.HOST` (default
      `http://localhost`) in its own `beforeEach`, which CLOBBERS `testEnvironmentOptions.url`. So in
      test bodies the origin comes from HOST, not from `url`. If you drop HOST, every test's origin
      silently falls back to `http://localhost` (not `:5601`). Keep both (`url` covers the
      pre-hook/non-mocked path), but HOST is the one that must be `:5601`.
- [ ] Re-declare non-configurable `localStorage`/`sessionStorage` as configurable.
- [ ] Replace `import '@testing-library/jest-dom/extend-expect'` →
      `import '@testing-library/jest-dom'` **everywhere** (setup file AND every test that
      imports it directly). The `/extend-expect` subpath was removed in jest-dom v6 and fails
      the whole suite to load. **Extremely common.**
- [ ] Add `TextEncoder`/`TextDecoder` polyfill via `require('util')` if any server-side test
      imports `react-dom/server` ("TextEncoder is not defined").
- [ ] If a setup file stubs `window.URL`, stub only the method
      (`window.URL.createObjectURL = () => ''`) — do NOT do `window.URL = {...}`, which destroys
      the `URL` constructor and breaks `new URL(...)` under jsdom 26.

### Toolchain — stale local Jest 27 shadow (STRUCTURAL BLOCKER, check first)
- [ ] If the plugin pins its own `jest-cli`/`jest-environment-jsdom@27` (check
      `node_modules/jest-environment-jsdom/package.json` version), the local copy shadows the
      parent's Jest 30 → every suite crashes with
      `Cannot read properties of undefined (reading 'html')`. Fix:
      `testEnvironment: '<rootDir>/../../node_modules/jest-environment-jsdom'` (absolute path).
      **Durable fix requires bumping the devDep pins in package.json — flag in the PR;** a
      `yarn install` would otherwise reintroduce the shadow.

### ESM-only dependencies (Jest 30 is stricter — the dominant new failure class)
- [ ] **`query-string` v9 (most common — reached via `opensearch_dashboards_react`/`utils` or
      core server mocks).** "Cannot use import statement outside a module". Three-part fix:
      (1) `moduleNameMapper` `'^query-string$': '<rootDir>/test/mocks/queryStringMock.js'`;
      (2) create that mock (require the real package by `__dirname`-relative absolute path, then
      normalize `.default`); (3) `transformIgnorePatterns` negative-lookahead adding
      `query-string|decode-uri-component|filter-obj|split-on-first`.
      NUANCE: if the plugin already uses a **curated** lookahead, just extend it (skip the mock).
- [ ] **Other pure-ESM transitive deps break the same way** — e.g. `jsonpath-plus` (+`jsep`,
      `@jsep-plugin/*`): add to the transform lookahead. Large trees like
      `@osd/monaco`/`monaco-editor`: add a global `jest.mock('@osd/monaco', ...)` mirroring
      `src/dev/jest/setup/monaco_mock.js` instead of transforming.
- [ ] **Jest 30 honors the `exports` field.** A dual `main`(UMD)+`exports`(ESM) dep (e.g.
      `@mapbox/mapbox-gl-draw`) now resolves to ESM and breaks — moduleNameMapper-pin it back to
      the prebuilt UMD dist bundle.
- [ ] Fixes are COUPLED: one ESM error masks the next. Re-run after each fix.

### Codemod (test files)
- [ ] `.toThrowError` → `.toThrow` (NOT `toThrowErrorMatchingInlineSnapshot`; perl
      `s/\.toThrowError(?!Matching)\b/.toThrow/g`).
- [ ] `.toBeCalledTimes`→`.toHaveBeenCalledTimes`, `.toBeCalledWith`→`.toHaveBeenCalledWith`,
      `.toBeCalled`→`.toHaveBeenCalled` (Times/With before the bare one).

### raw-loader
- [ ] Replace any `'^!!raw-loader!.*': 'jest-raw-loader'` mapper with a local stub
      `test/__mocks__/rawLoaderMock.js` (`module.exports = 'raw-loader-stub';`). jest-raw-loader
      is incompatible with the Jest 28+ transformer API.

### Snapshots
- [ ] Bulk-replace the header `https://goo.gl/fbAQLP` → `https://jestjs.io/docs/snapshot-testing`
      in ALL `*.snap`. Jest 30 REFUSES to load the old header — it fails the whole suite to RUN,
      massively inflating the apparent failure count.
- [ ] Then `-u --maxWorkers=2` to regenerate. **Manually review** non-header diffs
      (`git diff -- '*.snap' | grep '^[-+]' | grep -vE 'goo.gl|jestjs.io'`): confirm ONLY jsdom-26
      serialization changes (dropped `font-family: -webkit-small-control`, computed `calc()`
      values, `width: fit-content`, `aria-hidden` on `EuiIcon`, style reorder). Do NOT
      blind-accept genuine content/text/structure changes — investigate.

### window.location (jsdom 26 makes it non-configurable)
- [ ] Remove `delete window.location` and `Object.defineProperty(window, 'location', ...)` in
      tests (they throw). jest-location-mock provides `.assign/.replace/.reload` spies.
- [ ] SOURCE `window.location.href = X` → `window.location.assign(X)` — **REQUIRED, not optional**
      (the mock's href setter rejects relative URLs; `.assign()` accepts them). Grep
      `.location.href =` broadly (catches non-`window.` receivers like `thisWindow.location`).
- [ ] Anchor `<a href="#/x">` clicks NO LONGER navigate under jsdom 26. Assert the `href`
      attribute, or set the hash via `window.location.assign('#/x')` before render if a mount
      effect reads `window.location.hash`.
- [ ] NEVER replace the whole `window.history` — jest-location-mock spies on `history.pushState`.
      `jest.spyOn(window.history, 'replaceState')` on the real object instead.
- [ ] Fix cross-test location leakage exposed by per-test isolation.

---

## Plugin-specific cases
Beyond the common changes above, some plugins hit a smaller set of case-by-case fixes (enzyme
adapter swap, EUI/`userEvent` interaction changes, unawaited `findBy*`, modern-fake-timer
deadlocks, custom test runners, latent bugs unmasked by per-suite isolation, etc.). These do NOT
belong in this issue — **document whatever actually applied in the plugin PR description.** See the
reference migration (core PR opensearch-project/OpenSearch-Dashboards#12381) and the campaign recipe
for the full catalog if you need it.

## Acceptance criteria
- [ ] Full suite green (0 failed), exit 0.
- [ ] Plugin `tsc --noEmit` clean (ignoring pre-existing tsconfig warnings).
- [ ] Snapshot diffs reviewed and confirmed serialization-only.
- [ ] Note whether a `package.json` change is needed (stale Jest 27 pins) — most plugins inherit
      the parent's jest stack and need none. If the plugin pins its own `jest-cli`/
      `jest-environment-jsdom@^27`, drop `jest-cli` (bundled in jest 30) and bump
      `jest-environment-jsdom` to `^30.4.1` (and any `@types/jest` resolution to `^30.0.0`).
- [ ] If `package.json` changed, regenerate and commit the plugin's `yarn.lock` — run
      `yarn osd bootstrap --single-version=loose` from the OSD root, then include the updated
      `yarn.lock` in the PR. (Test-only PRs that don't touch `package.json` need no lockfile change.)

## Notes for reviewers
Changed files are dominated by `.snap` header bumps and the `extend-expect`/`query-string` fixes —
the substantive review is the `window.location` test edits, any ESM-transform config, and any
non-header snapshot diffs.

SuZhou-Joe:
Closes #1225

### Description

Migrate this plugin's Jest test suite to run under **Jest 30.4.2 + jsdom 26.1.0**, following the
core OpenSearch-Dashboards upgrade in opensearch-project/OpenSearch-Dashboards#12381. The plugin
runs its tests against the parent repo's jest binary, so it breaks once the core upgrade lands.

**Result:** full suite green (deterministic serial run) — `131 suites passed / 964 tests passed /
88 snapshots passed` (1 suite + 2 tests intentionally skipped), exit 0.

> Fixed a parallel-run flake that Jest 30 exposed: `DailyInsights/utils/__tests__/agentTaskStorage.test.tsx`
> declared `jest.mock('.../src/core/public/utils', …, { virtual: true })` for a module that actually
> exists. Under Jest 30's stricter resolver the virtual mock races with the real module across
> workers and intermittently fails to apply (`mountReactNode` un-mocked → 0 calls), which also
> polluted whichever suite shared the worker (seen as sporadic `List.test.tsx` failures). Removing
> `{ virtual: true }` makes it a normal mock of the real module; verified green across 6 consecutive
> full parallel runs.

#### Common changes (shared across the plugin-migration campaign)
- **Config (`test/jest.config.js`):** added `jest-location-mock` as the first `setupFilesAfterEnv`
  entry; added `testEnvironmentOptions: { url: 'http://localhost:5601' }`; added the
  `snapshotFormat: { escapeString: true, printBasicPrototype: true }` compatibility shim (Jest 29
  flipped these defaults). Also consolidated a duplicate trailing `setupFiles` block that had been
  silently overriding the polyfill/setup entries, folding `jest-canvas-mock` into the single
  `setupFiles` array.
- **Setup (`test/setup.jest.ts`):** set `process.env.HOST = 'http://localhost:5601'` so
  jest-location-mock's `window.location` base URL matches `testEnvironmentOptions.url`; re-declared
  the non-configurable `localStorage`/`sessionStorage` as configurable (jsdom 26 marks them
  non-configurable, which broke per-test `Object.defineProperty` overrides); fixed the jest-dom
  import from the removed `@testing-library/jest-dom/extend-expect` subpath to
  `@testing-library/jest-dom`.
- **Snapshots:** bulk-updated the snapshot header URL (`goo.gl/fbAQLP` →
  `jestjs.io/docs/snapshot-testing`) across 51 `.snap` files — Jest 30 refuses to load the old
  header. Every non-header content diff is jsdom-26 CSS serialization only (dropped
  `font-family: -webkit-small-control` on hidden mirror-`<div>`s used by auto-sizing inputs), no
  content regressions.

#### Plugin-specific changes
- **`window.location` (jsdom 26 makes it non-configurable):** changed the one source site that
  reassigned `window.location.href` to `window.location.assign(...)`
  (`public/pages/ForecastersList/utils/ForecastPopover.tsx` — "View forecast" button).
- **Removed obsolete `window.location` stubs from 11 test files:** each had a
  `beforeAll(() => Object.defineProperty(window, 'location', { value: {...}, writable: true }))`
  block that throws under jsdom 26 (location is non-configurable). jest-location-mock now provides
  a working `window.location`, so the manual stubs were deleted:
  `RefreshForecastersButton`, `AnomalyCharts/.../AnomaliesChart`, `AnomalyDetailsChart`,
  `AnomalyOccurrenceChart`, `DailyInsights/containers/DailyInsights`, `DailyInsights/utils/discoverLink`,
  `Dashboard/Components/AnomaliesLiveCharts`, `DetectorsList/.../EmptyMessage`,
  `ForecastersList/.../EmptyMessage`, `Overview/.../SampleDataBox`, and
  `Overview/containers/AnomalyDetectionOverview`.
- **`discoverLink.test.tsx`:** dropped the hand-rolled `window.location` origin/pathname stub and
  rely on `testEnvironmentOptions.url` (origin `http://localhost:5601`, pathname `/`) for the URL
  assertions.
- **number-input tests (`ConfigureModel/.../Settings.test.tsx`):** jsdom 26 sanitizes
  `type=number` inputs, so `userEvent.type(field, '-1')` drops the intermediate `-` and leaves
  `1`. Switched those cases to `fireEvent.change(field, { target: { value: '-1' } })` to preserve
  the negative-value intent.
- **async `findBy*` assertions (`ConfigureModel/.../Settings.test.tsx`):** fixed
  `expect(findByText(...)).not.toBeNull()` (which asserts on an unawaited Promise, always truthy)
  to `expect(await findByText(...)).toBeInTheDocument()` so the validation-error assertions are
  meaningful.

No `package.json` change needed — the plugin inherits the parent repo's Jest 30 stack.

### Issues Resolved

Part of the Jest 30 + jsdom 26 migration campaign (core: opensearch-project/OpenSearch-Dashboards#12381).

### Check List
  - [x] All tests pass, including unit test, integration test and doctest
  - [ ] New functionality has javadoc added
  - [ ] New functionality has user manual doc added
- [ ] Commits are signed per the DCO using --signoff

### DIFF
--- a/public/components/CodeModal/__tests__/__snapshots__/CodeModal.test.tsx.snap
+++ b/public/components/CodeModal/__tests__/__snapshots__/CodeModal.test.tsx.snap
@@ -1,4 +1,4 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
+// Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
 
 exports[`CodeModal spec renders the component 1`] = `
 <div
--- a/public/components/ContentPanel/__snapshots__/ContentPanel.test.tsx.snap
+++ b/public/components/ContentPanel/__snapshots__/ContentPanel.test.tsx.snap
@@ -1,4 +1,4 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
+// Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
 
 exports[`<ContentPanel /> spec renders the component 1`] = `
 <div
--- a/public/components/FeatureAnywhereContextMenu/AssociatedDetectors/components/__tests__/__snapshots__/EmptyAssociatedDetectorMessage.test.tsx.snap
+++ b/public/components/FeatureAnywhereContextMenu/AssociatedDetectors/components/__tests__/__snapshots__/EmptyAssociatedDetectorMessage.test.tsx.snap
@@ -1,4 +1,4 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
+// Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
 
 exports[`ConfirmUnlinkDetectorModal spec renders the component with filter applied 1`] = `
 <div>
--- a/public/components/RefreshForecastersButton/__tests__/RefreshForecastersButton.test.tsx
+++ b/public/components/RefreshForecastersButton/__tests__/RefreshForecastersButton.test.tsx
@@ -24,15 +24,6 @@ jest.mock('../../../services', () => ({
 
 describe('<RefreshForecastersButton /> spec', () => {
   beforeAll(() => {
-    Object.defineProperty(window, 'location', {
-      value: {
-        href: 'http://test.com',
-        pathname: '/',
-        search: '',
-        hash: '',
-      },
-      writable: true
-    });
   });
 
   beforeEach(() => {
--- a/public/components/RefreshForecastersButton/__tests__/__snapshots__/RefreshForecastersButton.test.tsx.snap
+++ b/public/components/RefreshForecastersButton/__tests__/__snapshots__/RefreshForecastersButton.test.tsx.snap
@@ -1,4 +1,4 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
+// Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
 
 exports[`<RefreshForecastersButton /> spec renders component 1`] = `
 <div
--- a/public/pages/AnomalyCharts/components/AlertsButton/__tests__/__snapshots__/AlertsButton.test.tsx.snap
+++ b/public/pages/AnomalyCharts/components/AlertsButton/__tests__/__snapshots__/AlertsButton.test.tsx.snap
@@ -1,4 +1,4 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
+// Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
 
 exports[`<AlertsButton /> spec Alerts Button renders component with monitor 1`] = `
 <div>
--- a/public/pages/AnomalyCharts/components/AlertsFlyout/__tests__/__snapshots__/AlertsFlyout.test.tsx.snap
+++ b/public/pages/AnomalyCharts/components/AlertsFlyout/__tests__/__snapshots__/AlertsFlyout.test.tsx.snap
@@ -1,4 +1,4 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
+// Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
 
 exports[`<AlertsFlyout /> spec Alerts Flyout renders component with monitor 1`] = `
 <div
--- a/public/pages/AnomalyCharts/components/AnomaliesStat/__tests__/__snapshots__/AnomalyStat.test.tsx.snap
+++ b/public/pages/AnomalyCharts/components/AnomaliesStat/__tests__/__snapshots__/AnomalyStat.test.tsx.snap
@@ -1,4 +1,4 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
+// Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
 
 exports[`<AlertsStat /> spec Alert Stat renders component with monitor and loading 1`] = `
 <div>
--- a/public/pages/AnomalyCharts/containers/__tests__/AnomaliesChart.test.tsx
+++ b/public/pages/AnomalyCharts/containers/__tests__/AnomaliesChart.test.tsx
@@ -65,15 +65,6 @@ const renderDataFilter = (chartProps: AnomaliesChartProps) => ({
 
 describe('<AnomaliesChart /> spec', () => {
   beforeAll(() => {
-    Object.defineProperty(window, 'location', {
-      value: {
-        href: 'http://test.com',
-        pathname: '/',
-        search: '',
-        hash: '',
-      },
-      writable: true
-    });
   });
   test('renders the component for sample / preview', () => {
     console.error = jest.fn();
--- a/public/pages/AnomalyCharts/containers/__tests__/AnomalyDetailsChart.test.tsx
+++ b/public/pages/AnomalyCharts/containers/__tests__/AnomalyDetailsChart.test.tsx
@@ -66,15 +66,6 @@ const renderAnomalyOccurenceChart = (
 
 describe('<AnomalyDetailsChart /> spec', () => {
   beforeAll(() => {
-    Object.defineProperty(window, 'location', {
-      value: {
-        href: 'http://test.com',
-        pathname: '/',
-        search: '',
-        hash: '',
-      },
-      writable: true
-    });
   });
   
   test('renders the component in case of Sample Anomaly', () => {
--- a/public/pages/AnomalyCharts/containers/__tests__/AnomalyOccurrenceChart.test.tsx
+++ b/public/pages/AnomalyCharts/containers/__tests__/AnomalyOccurrenceChart.test.tsx
@@ -67,15 +67,6 @@ const renderAnomalyOccurenceChart = (
 
 describe('<AnomalyOccurrenceChart /> spec', () => {
   beforeAll(() => {
-    Object.defineProperty(window, 'location', {
-      value: {
-        href: 'http://test.com',
-        pathname: '/',
-        search: '',
-        hash: '',
-      },
-      writable: true
-    });
   });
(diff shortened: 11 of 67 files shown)


SuZhou-Joe (commented):
Changes reviewed, all related to jest change except replace window.location.href = -> window.location.assign()

ohltyler (approved):

PR #1238, opened 2026-08-06
Integrate centralized resource-sharing share button for anomaly-detector and forecaster

DarshitChanpura:
### Description

Integrates the centralized share button from [security-dashboards-plugin PR #2491](https://github.com/opensearch-project/security-dashboards-plugin/pull/2491) via its dependency-free DOM-marker SPI.

**How it works**
- Renders empty marker elements (`data-resource-share-button` + resource id/type attributes); the security plugin mounts its Share/Update Access button + modal into them when installed and resource sharing is enabled
- UI chrome (Share column) is gated on the core `resourceSharing` capability (`enabled` + type present in `availableTypes`), so it is completely absent when the feature is off or the security plugin is missing
- No dependency on the security plugin: no manifest changes, no imports from it — markers are inert empty elements when unfulfilled

**Testing**: verified manually on OpenSearch 3.8.0 (staging) + OSD main with resource sharing enabled — share/update flows, `can_share: false` graying, and feature-off column hiding.

### Issues Resolved

### Check List

  - [ ] All tests pass


DarshitChanpura:
Detector:
Screenshot 2026-08-24 at 5 09 16 PM

Forecaster:
Screenshot 2026-08-24 at 5 10 13 PM

public/pages/ForecastersList/containers/List/List.tsx
```diff
@@ -816,7 +816,7 @@ const handleStartForecasterJob = async (forecasterId: string, forecasterName: st
                  * row heights and truncates longer text values appropriately.
                  */
                 rowHeightsOptions={{
-                  defaultHeight: { lineCount: 1 },
+                  defaultHeight: 'auto',
```
kaituo: 'auto' is unsupported by the current OUI version (https://github.com/opensearch-project/oui/blob/1.22.1/src/components/datagrid/data_grid_types.ts#L344-L356). It falls back to the minimum height while removing one-line clamping, so content may wrap and clip.

public/pages/utils/helpers.ts
```diff
@@ -422,3 +423,23 @@ export const mapToVisibleForecasterOptions = (items: any[], key: string) =>
...
+ * core capability registered by security-dashboards-plugin. False when that
+ * plugin is not installed, the feature is disabled, or the type is not
+ * registered with the resource-sharing framework — no plugin dependency
+ * involved.
+ */
+export function isResourceSharingAvailable(
+  resourceType: string = AD_RESOURCE_TYPE
+): boolean {
+  try {
+    const caps = (getApplication().capabilities as any)?.resourceSharing;
```
kaituo: Resource sharing is a backend, cluster/collection-level setting, correct? With MDS, the same OpenSearch Dashboards instance can connect to data sources running different versions or in AOS/AOSS, so some may support resource sharing while others do not.

This code gates the UI using `getApplication().capabilities.resourceSharing`, which is resolved from local Dashboard setting, while the mounted button targets the selected remote dataSourceId. Is resource-sharing availability intended to be global across all data sources in a deployment, or should it be evaluated per selected data source? If it is data-source-specific, this could leave an empty Access column or hide valid sharing controls.

public/pages/ForecastersList/containers/List/List.tsx
DarshitChanpura: Good catch — this was an unintended change on my part. Reverted to `defaultHeight: { lineCount: 1 }` in 598c077, which is supported by the OUI version in use and matches the comment's intent (consistent single-line, truncated rows). `List.tsx` now has no net change from `main`.

public/pages/utils/helpers.ts
DarshitChanpura: Good question. Yes, resource sharing is a backend, per-cluster feature, so it's per data source — and the integration reflects that at the operation level:

- The marker forwards the selected data source: the Access column renders `data-resource-data-source-id={dataSourceId}` when a data source is selected.
- The centralized share button reads that attribute (`share-button-dom-spi.tsx`) and builds its client with `buildResourceApi(http, dataSourceId)`, so all list/share/update operations target the **selected** data source's backend, not the local cluster.

The one part that is not per-source today is the *column visibility* gate. `getApplication().capabilities.resourceSharing` is resolved by the security-dashboards-plugin capability switcher from the cluster the capabilities request authenticates against (the local/default connection), so in a mixed MDS deployment the column's visibility follows the local cluster while the per-row operations correctly target the selected source.

Making the column-visibility gate itself per-`dataSourceId` would be a change in the capability layer that security-dashboards-plugin owns (it registers the `resourceSharing` capability and mounts the button), rather than in this plugin — the AD integration already forwards the data source correctly. I can open a follow-up issue on security-dashboards-plugin to resolve the capability per selected data source if that's the desired behavior for heterogeneous deployments.

kaituo: yeah, please fix the visibility gate issue. Would gate the UI via the response of buildResourceApi(http, dataSourceId) be easier? I would prefer the gate issue fixed before merging the PR. AD supports AOS and AOSS and a lot of (>50%) versions of AOS users still use versions before 1.3. AOSS AD does not support resource sharing at all.

DarshitChanpura: Fixed in 48a95de — the Access column is now gated per selected data source rather than on the local Dashboards capability.

- New AD server route `.../resource_sharing_availability/{resourceType}/{dataSourceId}` probes the **selected** data source via `getClientBasedOnDataSource(...).transport.request` to `GET /_plugins/_security/api/resource/types` and returns `{ available: types.includes(resourceType) }`.
- It **fails closed**: any error (feature disabled, endpoint absent on older AOS versions, or AOSS/serverless where the security resource API doesn't exist) resolves to `available: false`, so the column is hidden for data sources that don't support resource sharing.
- The Detectors and Forecasters lists (and the detail pages) resolve availability asynchronously with a `useEffect` keyed on the selected data source, so switching data sources re-evaluates. This matches the pattern already used for `isServerlessDataSource`.
- I removed the local-capability `isResourceSharingAvailable` helper entirely and updated the unit tests (fail-closed cases + per-source gating).

This aligns column visibility with the per-row share operations, which already targeted the selected data source via `buildResourceApi(http, dataSourceId)`. So AOSS and pre-resource-sharing AOS data sources no longer show an empty/non-functional Access column.

public/pages/ForecastersList/containers/List/List.tsx
```diff
@@ -580,7 +608,7 @@ export const ForecastersList = (props: ListProps) => {
     }, [getSavedObjectsClient(), getNotifications(), props.setActionMenu]);
   }
 
-  const columns = getDataGridColumns();
+  const columns = getDataGridColumns(resourceSharingAvailable);
```
kaituo, line 611: columns is only initialized once in

```
const [visibleColumns, setVisibleColumns] = useState(
    columns.map(({ id }) => id) // initialize to the full set of columns
  );
```

When users change data source, resourceSharingAvailable is changed, but useState won't reload visibleColumns.

server/routes/ad.ts
```diff
@@ -216,6 +226,58 @@ export function registerADRoutes(apiRouter: Router, adService: AdService) {
...
+        return opensearchDashboardsResponse.ok({
+          body: { ok: true, available: false },
+        });
+      }
+
+      // Per-type gate: the resource type must be a registered/protected shareable
+      // type (admins can enable sharing for only a subset of types).
+      const response = await callWithRequest('transport.request', {
+        method: 'GET',
+        path: '/_plugins/_security/api/resource/types',
```
kaituo, line 265: From https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/resources/api/list/ResourceTypesRestAction.java#L73, the api would return ad or forecasting enabled as long as the AD plugin version starts supporting resource sharing, right? You are not checking if resource sharing is enabled via a dynamic setting.

kaituo (commented):
haven't finished reading new files. Will continue later.

public/pages/ForecastersList/containers/List/List.tsx
DarshitChanpura: Fixed in 6ceb1a8. The forecasters list uses `EuiDataGrid`, whose `visibleColumns` state was seeded once from `columns` on mount, so the Access column didn't refresh when the data source (and thus resource-sharing availability) changed. Added a `useEffect` that re-syncs it to the current column set:

```ts
useEffect(() => {
  setVisibleColumns(
    getDataGridColumns(resourceSharingAvailable).map(({ id }) => id)
  );
}, [resourceSharingAvailable]);
```

(The detectors list uses `EuiBasicTable`, which recomputes columns each render, so it wasn't affected.)

server/routes/ad.ts
DarshitChanpura: Addressed in 6ceb1a8 by converging AD onto the same front-end gating helper the other resource-sharing consumer plugins use (reporting, notifications, security-analytics, ml-commons, flow-framework) instead of the bespoke server route. `getResourceSharingAvailableTypes(dataSourceId)` now:

1. `GET /api/v1/auth/resource_sharing_enabled` — the security-dashboards-plugin endpoint backed by `dashboardsinfo.resource_sharing_enabled` (the dynamic-setting gate); returns `[]` when not enabled.
2. `GET /api/resource/types` — the registered/protected type list.

So the enabled-setting check is now the first gate, matching the rest of the ecosystem, and the bespoke `/resource_sharing_availability` route is removed. (For the record, the old route did already gate on `dashboardsinfo.resource_sharing_enabled` before the types call — but converging on the shared helper keeps AD consistent with the other five plugins and reuses the same endpoint added in security-dashboards-plugin#2520.)

public/pages/DetectorsList/containers/List/List.tsx
```diff
@@ -177,6 +178,12 @@ export const DetectorList = (props: ListProps) => {
 
   const [localClusterName, setLocalClusterName] = useState("");
 
+  // Whether resource sharing is available on the SELECTED data source. Gates
+  // the Access column per data source (a backend setting), rather than the
+  // local Dashboards capability. Defaults to false and fails closed.
```
kaituo: Do we need both local Dashboards capability and backend setting to be true here? We can have a compatible data source, but if security dashboard is not installed or local Dashboards's resource sharing capability is off, resourceSharingAvailable is false.

public/pages/utils/helpers.ts
```diff
@@ -422,3 +422,39 @@ export const mapToVisibleForecasterOptions = (items: any[], key: string) =>
...
+ * flow-framework). Returns [] when resource sharing is disabled or on any error
+ * (fails closed).
+ */
+export async function getResourceSharingAvailableTypes(
+  dataSourceId?: string
+): Promise<string[]> {
+  try {
+    const query =
+      dataSourceId && dataSourceId.trim().length > 0 ? { dataSourceId } : {};
+    // Global gate: resource sharing must be enabled on the selected data source.
```
kaituo, line 441: This function checks resource-sharing enablement and registered resource types on the selected backend/data source, but does not verify that the Security Dashboards client-side DOM-marker SPI is running (https://github.com/opensearch-project/security-dashboards-plugin/blob/main/public/plugin.ts#L145), right? If SPI is not enabled and data source is enabled, we would get an empty column.

public/pages/ForecastersList/containers/List/List.tsx
```diff
@@ -614,6 +635,16 @@ export const ForecastersList = (props: ListProps) => {
     columns.map(({ id }) => id) // initialize to the full set of columns
   );
 
+  // Re-sync the visible column set when the Access column is added or removed.
```
kaituo: If the user hides Status or Indices, resolving availability or switching sources makes them visible again. Only add/removes for existing rows.

public/pages/utils/helpers.ts
DarshitChanpura: Fixed in 48a95de — the Access column is now gated per selected data source rather than on the local Dashboards capability.

- New AD server route `.../resource_sharing_availability/{resourceType}/{dataSourceId}` probes the **selected** data source via `getClientBasedOnDataSource(...).transport.request(...)` against the security plugin's per-type check.
- The front-end helper calls this route with the selected `dataSourceId` instead of reading the local `resourceSharingAvailable` capability.

Follow-up on 09-14: you're right that this still didn't confirm the client-side DOM-marker SPI is actually running, only that the *selected data source* reports sharing enabled/registered. Fixed properly in 814894c: `getResourceSharingAvailableTypes` now declares `securityDashboards` as an optional dependency and re-confirms each backend-reported type via `securityDashboards.ui.isResourceSharingAvailable(type, dataSourceId)` before including it — failing closed to `[]` entirely when security-dashboards-plugin isn't installed. That method itself is now fixed to fail closed on the *local* SPI in opensearch-project/security-dashboards-plugin#2525, so the exact scenario you described (local cluster has sharing disabled, selected data source has it enabled) now correctly hides the column instead of rendering it with a dead Share button. Added tests exercising this in 814894c.

public/pages/ForecastersList/containers/List/List.tsx
DarshitChanpura: Good catch — fixed in 814894c. The `useEffect` reset `visibleColumns` to the full `columns` set on every availability change, so any column the user had manually hidden (Status, Indices, etc.) reappeared. It now adds/removes only the `'share'` id from the current `visibleColumns` via a functional `setVisibleColumns` update, leaving everything else untouched.

Added a regression test that hides the `state` column, resolves resource-sharing availability, and asserts `state` stays hidden while `share` is added.

DarshitChanpura:
### Re: Code-Diff-Analyzer flag on `opensearch_dashboards.json`

The analyzer flagged the new `"securityDashboards"` entry in `optionalPlugins` per its mandatory "flag all new dependency changes" rule — this is expected, not a defect.

**For maintainer review:**
- `securityDashboards` resolves to the in-repo `security-dashboards-plugin` (relative import, e.g. `../../security-dashboards-plugin/public/types`) — not an npm package, so there is no typosquatting/namespace-hijack surface.
- It is declared in `optionalPlugins`, not `requiredPlugins`, matching the existing pattern already used for `assistantDashboards`/`dataSource`/etc. in this manifest.
- The dependency is optional at every level: an optional TypeScript field (`securityDashboards?: SecurityPluginStart`), and a runtime `if (securityDashboards)` guard in `start()`. When security-dashboards-plugin is absent or disabled, that block never runs and the resource-sharing helper fails closed to `[]` without ever calling into it — covered by a dedicated unit test (`returns [] without probing when security-dashboards-plugin is not installed`).
- The same dependency + pattern is added identically across the sibling PRs below, so a single review here covers the shape of the change everywhere it appears.

Once reviewed, the standard unblock is: add the `skip-diff-analyzer` label, then re-run the failed check.

public/pages/utils/helpers.ts
DarshitChanpura: You're right, and thanks for pinpointing this precisely. The fix landed at the actual root cause rather than in this helper.

security-dashboards-plugin's `startShareButtonDomSpi` — the DOM-marker watcher that mounts the Share button — was itself gated on the local cluster's resource-sharing flag, so it never started at all when local was disabled, even if the selected data source had sharing enabled. That's fixed in opensearch-project/security-dashboards-plugin#2525: the watcher now always starts, and each mounted button independently checks availability for its own `dataSourceId` (already correctly implemented) before rendering.

I initially tried gating this helper on `securityDashboards.ui.isResourceSharingAvailable` instead (i.e. re-checking the local flag here too), but that doubled down on the wrong coupling and introduced a new regression in the other direction: the Access column would incorrectly disappear when local is disabled but the selected data source has sharing enabled. That approach has been reverted from this PR — this helper is back to its original per-data-source logic, unchanged, which was already correct. No changes are needed in this plugin for the empty-column issue; opensearch-project/security-dashboards-plugin#2525 is the fix.

public/pages/utils/helpers.ts
DarshitChanpura: Following up here — fixed, but at the root cause rather than by gating this helper via `buildResourceApi(http, dataSourceId)` or similar.

The actual bug was in security-dashboards-plugin: its Share-button DOM-marker SPI watcher only started when the local cluster had resource sharing enabled, so no button could mount anywhere in the app if local was disabled, regardless of the selected data source's own setting. That's fixed in opensearch-project/security-dashboards-plugin#2525 — the watcher now always starts, and per-data-source availability (this helper, unchanged) correctly decides each button's visibility.

I'd initially added a redundant local-flag re-check to this helper too, which caused the opposite problem (column disappearing when it shouldn't). That's been reverted — see the reply on the other thread for details. This plugin needs no changes for the empty-column issue you raised.

public/pages/DetectorsList/containers/List/List.tsx
```diff
@@ -256,6 +263,22 @@ export const DetectorList = (props: ListProps) => {
...
+        if (!cancelled) setResourceSharingAvailableTypes(types);
+      }
+    );
+    return () => {
+      cancelled = true;
+    };
+  }, [state.selectedDataSourceId]);
+
+  const resourceSharingAvailable =
+    resourceSharingAvailableTypes.includes(AD_RESOURCE_TYPE);
```
kaituo: you missed my comment: https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/commit/7d070b317fefd0345c757ea93b4c6cf62ba0e834#r200266620

DarshitChanpura: Restored in a82dbef, along with the same guard in ForecastersList/List.tsx (which had the identical pattern dropped in the same commit), plus direct unit tests covering all 4 branches of the guard.

For completeness: I checked and this same race (async availability probe, no check that the resolved result still matches the currently-selected data source) exists in all of the other resource-sharing consumer plugins too (ml-commons-dashboards, security-analytics-dashboards-plugin, dashboards-reporting, dashboards-notifications, dashboards-flow-framework, alerting-dashboards-plugin) -- none of them had this guard to begin with. Adding it there as well across their respective PRs.

kaituo (approved):

DarshitChanpura:
Follow-up for visibility: the same data-source staleness guard has now been added to the resource-sharing Access column in all other consumer plugins, since none of them had it either. All are open and ready for review:

- ml-commons-dashboards#517 (merged)
- security-analytics-dashboards-plugin#1575
- dashboards-reporting#808
- dashboards-notifications#487
- dashboards-flow-framework#917
- alerting-dashboards-plugin#1512

The root-cause SPI fix in security-dashboards-plugin#2525 has also merged.

